PR #3, opened 2021-09-17
Add universal x86_64-linux static+shared build

### LINKED ISSUE #2: Align x86_64 "universal" linux build with aarch64
There's an asymmetry with `x86_64`:

- there's `libddwaf-1.0.11-linux-x86_64-static.tar.gz` which is a static lib that should work on alpine and debian but that I can't use for the Ruby bindings since I'm using FFI which needs a shared lib
- there's `libddwaf-1.0.11-linux-x86_64-muslc.tar.gz` and `libddwaf-1.0.11-linux-x86_64-glibc.tar.gz` which do have shared libs but don't run on both, and should be fine but are a bit awkward as to being different for another cpu arch but the difference is in fact unrelated to the cpu arch, just that it turns out it's using a different build process.

I have the feeling that `libddwaf-1.0.11-linux-x86_64-static.tar.gz` should drop the `-static` from the name and reuse/follow as much as possible the build process of `libddwaf-1.0.11-linux-aarch64.tar.gz`, resulting in a static+shared lib build for `x86_64`, uniform with `aarch64`. The build should be very similar expect without the ARM toolchain, so probably using the regular compiler instead of the cross compiling one should be fine, and then tests would run without qemu.

We could duplicate the `aarch64` dir as `x86_64` and proceed with the adjustments, to "make it work" and surface possible commonalities. A future step would look into sharing stuff to "make it right".

lloeki:
Closes #2


cataphract (dismissed):

Anilm3:
This new change now results in two libc++-static.tar.gz (one for `aarch64` and one for `x86_64`), I believe @Julio-Guerra  uses these so ideally can you rename them so that it's clear which is which. Also, the artefacts produced are only tar.gz instead of both tar.gz and zip. I think that should be fine but if no one is using the zip files then we might as well remove them altogether from all builds.

Anilm3 (changes requested):
This new change now results in two `libc++-static.tar.gz` (one for `aarch64` and one for `x86_64`), I believe @Julio-Guerra uses these so ideally can you rename them so that it's clear which is which. Also, the artefacts produced are only `tar.gz` instead of both `tar.gz` and `zip`. I think that should be fine but if no one is using the `zip` files then we might as well remove them altogether from all builds.

Julio-Guerra:
> This new change now results in two `libc++-static.tar.gz` (one for `aarch64` and one for `x86_64`), I believe @Julio-Guerra uses these so ideally can you rename them so that it's clear which is which. Also, the artefacts produced are only `tar.gz` instead of both `tar.gz` and `zip`. I think that should be fine but if no one is using the `zip` files then we might as well remove them altogether from all builds.

I would even put them in the libddwaf tarball because the library cannot be used without them.

docker/libddwaf/gcc/linux/Dockerfile
```diff
@@ -1,166 +0,0 @@
...
-# and C libraries. It has also the nice ability to provide a static unwinding
-# library, which is not on glibc (libgcc_s.so only).
-#
-# The resulting libddwaf and libc++ libraries are therefore able to work on
-# any GNU Linux distribution such as Alpine, CentOS, etc.
-#
-# So the requirements to compile libddwaf and libc++ are:
-#   - gcc and g++ for the linux target, ideally modern-enough to compile the
-#     standard C++ version of libddwaf and the libc++ for it.
-#   - musl library headers.
```
Julio-Guerra, line 18: I think this comment is worth keeping it somewhere. This took me quite some time to dig into and not everyone knows about it.

lloeki: This kind of detail should probably go in a scoped `docker/libddwaf/README.md` or general `BUILD.md` or something to that effect.

lloeki:
> libc++-static.tar.gz (one for aarch64 and one for x86_64)

yeah `libc++-static-aarch64-linux.tar.gz` and `libc++-static-x86_64-linux.tar.gz` sound good (or maybe `-static` at the end

> I would even put them in the libddwaf tarball because the library cannot be used without them.

I would keep them separate, as it highlights the dependency better, and could be used by third parties to great effect (I might try for libv8-node), and downloading two files instead of one isn't that hard. Coming from the libv8-node game, I would have really enjoyed Node.js to provide the libv8 monolith as an artifact of their node release builds, which would be very low maintenance for them, yet they don't; the situation is kind of similar.

Anilm3 (approved):
lgtm

PR #4, opened 2021-09-20
Function to return required addresses

Anilm3:


include/ddwaf.h
```diff
@@ -188,6 +188,18 @@ ddwaf_handle ddwaf_init(const ddwaf_object *rule, const ddwaf_config* config);
  */
 void ddwaf_destroy(ddwaf_handle handle);
 
+/**
+ * ddwaf_required_addresses
+ *
+ * Get a list of required (root) addresses. The memory is owned by the user, each
+ * element and the array should be freed with 'free'.
```
cataphract, line 195: Having the client of the library free memory allocated by the library is generally not a good idea, because there is no guarantee that the free inside the library was resolved to the same free in the client. In practice, this is mostly a problem on windows.

In this case, this problem can be easily side-stepped and a more friendly interface provided. Just make the memory owned be object pointed to by `ddwaf_handle`. This object can be a `std::vector<const char *>` whose elements are obtained by `std::string::c_str`, which is NUL-terminated. Then the C interface can just call `data()` on the vector. The result will be valid for the lifetime of the ddwaf_handle, since the manifest shouldn't change.

include/ddwaf.h
```diff
@@ -191,15 +191,15 @@ void ddwaf_destroy(ddwaf_handle handle);
...
+ * should not be freed.
  *
  * @param Handle to the WAF instance.
  * @param size Output parameter in which the size will be returned. The value of
- *             size will be undefined if the return value is NULL.
+ *             size will be 0 if the return value is nullptr.
  * @return NULL if error, otherwise a pointer to an array with size elements.
  **/
-char** ddwaf_required_addresses(const ddwaf_handle handle, uint32_t *size);
+const char** ddwaf_required_addresses(const ddwaf_handle handle, uint32_t *size);
```
cataphract, line 202: const char *const *

src/PWManifest.h
```diff
@@ -59,7 +60,7 @@ class PWManifest
     void insert(std::string_view name, ArgDetails &&arg);
     bool empty() { return argIDTable.empty(); }
 
-    std::vector<std::string_view> get_root_addresses();
+    std::vector<const char *>& get_root_addresses() { return root_addresses; };
```
cataphract, line 63: const std::vector<const char *>& get_root_addresses() const { return root_addresses; };

src/PowerWAFInterface.cpp
```diff
@@ -89,36 +89,18 @@ extern "C"
         }
     }
 
-    char** ddwaf_required_addresses(const ddwaf_handle handle, uint32_t *size) {
+    const char** ddwaf_required_addresses(const ddwaf_handle handle, uint32_t *size) {
```
cataphract, line 92: const char *const *

cataphract (approved):

PR #5, opened 2021-09-20
Fix an error in corpus generator

cbeauchesne:
* Use the good naming for `addresses`
* fix a small typo in rules generator, it should speed up the time-to-asymptote.


Anilm3 (approved):
lgtm

PR #8, opened 2021-09-21
Cross-compile to arm64-darwin on x86_64

### LINKED ISSUE #7: Add cross-compilation to arm64-darwin CI
This can be achieved on `x86_64` macOS runners by adding `-arch arm64` to compiler and linker flags.

Executing the tests is another matter but at least the builds would be automated.

lloeki:
Closes #7

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,32 @@ jobs:
         working-directory: ${{ github.workspace }}/packages
       - uses: actions/upload-artifact@v2
         with:
-          name: libddwaf-macos
+          name: libddwaf-macos-x86_64
+          path: ${{ github.workspace }}/packages
+
+  macos-cross-build:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Create Build Directory
+        run: cmake -E make_directory ${{ github.workspace }}/build ${{ github.workspace }}/packages
+      - name: Generating Build Scripts
+        run: cmake -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=${{ github.workspace }} -DCPACK_PACKAGE_DIRECTORY=${{ github.workspace }}/packages ${{ github.workspace }}
+        working-directory: ${{ github.workspace }}/build
+      - name: Build Binaries
+        run: cmake --build . --config RelWithDebInfo --verbose --target all -j
+        working-directory: ${{ github.workspace }}/build
+      - name: Build Packages
+        run: cmake --build . --target package --config RelWithDebInfo --verbose
+        working-directory: ${{ github.workspace }}/build
+      - name: Remove Temporary Files
+        run: cmake -E remove_directory _CPack_Packages
+        working-directory: ${{ github.workspace }}/packages
+      - uses: actions/upload-artifact@v2
+        with:
+          name: libddwaf-macos-arm64
           path: ${{ github.workspace }}/packages
 
   docker-builds:
@@ -165,7 +190,7 @@ jobs:
 
           extract_file ./artifacts/libddwaf-linux-x86_64/libddwaf-*-linux-x86_64*.tar.gz \
             '/libddwaf\.so$' ./packages/nuget/runtimes/linux-x64/native
-          extract_file ./artifacts/libddwaf-macos/libddwaf-*-darwin-x86_64*.zip \
+          extract_file ./artifacts/libddwaf-macos-x86_64/libddwaf-*-darwin-x86_64*.zip \
             '/libddwaf\.dylib$' ./packages/nuget/runtimes/osx-x64/native
           extract_file ./artifacts/libddwaf-windows-Win32/libddwaf-*-windows-win32*.zip \
             '/ddwaf\.dll$' ./packages/nuget/runtimes/win-x86/native
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,10 @@ file(READ "version" version)
 project(libddwaf VERSION ${version})
 
 set(CMAKE_OSX_DEPLOYMENT_TARGET "10.12" CACHE STRING "Minimum OS X deployment version")
+if(CMAKE_OSX_ARCHITECTURES MATCHES "x86_64" OR
+  CMAKE_OSX_ARCHITECTURES MATCHES "arm64")
+  set(CPU_TYPE ${CMAKE_OSX_ARCHITECTURES})
+endif()
 
 set(CMAKE_C_STANDARD 99)
 set(CMAKE_CXX_STANDARD 17)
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -144,16 +152,31 @@ if (LIBDDWAF_BUILD_STATIC)
...
+              COMMAND rm -rf ar_comb
+              WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+        else()
+          add_custom_command(TARGET libddwaf_static POST_BUILD
+              COMMAND ${CMAKE_COMMAND} -E make_directory ar_comb
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:libddwaf_static>
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:lib_re2>
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:lib_ac>
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:lib_injection>
+              COMMAND cd ar_comb && ${CMAKE_AR} -qcs combined${CMAKE_STATIC_LIBRARY_SUFFIX} *.o*
```
lloeki, line 174: When cross-compiling on macOS this produces a fat `x86_64` (because it's the native arch) + `arm64` (because there are `arm64` `.o`) archive, even though there is no `x86_64` `.o`. So either one has to use `lipo` to get only the `arm64` part as a thin archive, or just use `libtool` as above.

CMakeLists.txt
```diff
@@ -144,16 +152,31 @@ if (LIBDDWAF_BUILD_STATIC)
...
-            COMMAND rm -rf ar_comb
-            WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+        if(CMAKE_OSX_ARCHITECTURES MATCHES "arm64")
+          add_custom_command(TARGET libddwaf_static POST_BUILD
+              COMMAND ${CMAKE_COMMAND} -E make_directory ar_comb
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:libddwaf_static>
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:lib_re2>
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:lib_ac>
+              COMMAND ${CMAKE_COMMAND} -E chdir ar_comb ${CMAKE_AR} -x $<TARGET_FILE:lib_injection>
+              COMMAND cd ar_comb && libtool -static -o combined.a -arch_only ${CMAKE_OSX_ARCHITECTURES} *.o*
```
lloeki, line 162: So this case uses `libtool` with an additional flag. Maybe `libtool` could be used on all platforms instead of `ar`, which would help in getting rid of the big `if()`?

CMakeLists.txt
```diff
@@ -11,7 +15,11 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(LIBDDWAF_SHARED_LINKER_FLAGS "-static-libstdc++" CACHE STRING "Shared library extra linker flags")
 set(LIBDDWAF_EXE_LINKER_FLAGS "" CACHE STRING "Executable extra linker flags")
 set(LIBDDWAF_PACKAGE_SUFFIX "" CACHE STRING "Suffix for packaging purposes")
-set(LIBDDWAF_PACKAGE_PROCESSOR ${CMAKE_SYSTEM_PROCESSOR} CACHE STRING "Alternative processor for packaging purposes")
+if(CMAKE_OSX_ARCHITECTURES MATCHES "arm64")
```
lloeki, line 18: I don't particularly like that, but I'm not well versed enough in `cmake` to improve this significantly.

CMakeLists.txt
```diff
@@ -3,6 +3,10 @@ file(READ "version" version)
 project(libddwaf VERSION ${version})
 
 set(CMAKE_OSX_DEPLOYMENT_TARGET "10.12" CACHE STRING "Minimum OS X deployment version")
+if(CMAKE_OSX_ARCHITECTURES MATCHES "x86_64" OR
+  CMAKE_OSX_ARCHITECTURES MATCHES "arm64")
+  set(CPU_TYPE ${CMAKE_OSX_ARCHITECTURES})
```
lloeki, line 8: This seemed to be required early on, not sure if it still is with subsequent changes.

.github/workflows/build.yml
```diff
@@ -165,7 +190,7 @@ jobs:
 
           extract_file ./artifacts/libddwaf-linux-x86_64/libddwaf-*-linux-x86_64*.tar.gz \
             '/libddwaf\.so$' ./packages/nuget/runtimes/linux-x64/native
-          extract_file ./artifacts/libddwaf-macos/libddwaf-*-darwin-x86_64*.zip \
+          extract_file ./artifacts/libddwaf-macos-x86_64/libddwaf-*-darwin-x86_64*.zip \
```
lloeki, line 193: Maybe one should also add `libddwaf-macos-arm64` at some point, but I'm not sure if .Net supports it yet so I left it at `x86_64`.

.github/workflows/build.yml
```diff
@@ -66,7 +66,32 @@ jobs:
         working-directory: ${{ github.workspace }}/packages
       - uses: actions/upload-artifact@v2
         with:
-          name: libddwaf-macos
+          name: libddwaf-macos-x86_64
+          path: ${{ github.workspace }}/packages
+
+  macos-cross-build:
```
lloeki, line 72: This is mostly a dupe of `macos-build`, except:

- it adds a `cmake` variable to cross compile
- it skips test building: one less target on the build step; it does build but attempts to run the binary right after which makes it fail
- it skips the test running step altogether

Maybe that can be folded into `macos-build` with a matrix and a couple of variables + conditional steps.

lloeki:
I've downloaded and tested the `e54c38d` arm64 dylib build artifact on my M1 machine using the Ruby bindings and it works fine.

CMakeLists.txt
cataphract: Seems to do fine without it. Also `CPU_TYPE` is not something cmake knows about, so I don't know what effect this could have.

CMakeLists.txt
cataphract: I don't think there's a better alternative. `OSX_ARCHITECTURES` doesn't seem to influence other variables. It would be difficult too, since, as the name suggests, there can be more than one architecture.

CMakeLists.txt
cataphract: It's not pretty, but I think it's preferable to require that linux systems have `libtool` installed. Another option would be to try to get rid of the `else` by using `-Wl,--whole-archive`, which I don't think is supported in mac os.

cataphract (approved):

CMakeLists.txt
lloeki: I don't anticipate us building universal fat binaries, so I guess there would be only one ever for the foreseeable future.

lloeki:
Thanks, I'll merge this as is to have the artifact continuously available right away, and try to improve on the mentioned things progressively in the future.

PR #9, opened 2021-10-01
LibInjection XSS and SQL rule processors

Anilm3:


src/PWRet.cpp
```diff
@@ -76,8 +76,8 @@ void PWRetManager::recordRuleMatch(const std::unique_ptr<IPWRuleProcessor>& proc
     rapidjson::Value output, opNameValue, opValue, baValue, manifestKey;
     output.SetObject();
 
-    const std::string& opName = processor->operatorName();
-    opNameValue.SetString(opName.c_str(), static_cast<rapidjson::SizeType>(opName.size()));
+    std::string_view opName = processor->operatorName();
```
cataphract, line 79: why this change? are you relying on the std::string_view being backed by a nul-terminated static string? because that's generally not part of the contract of std::string_view and may cause confusion/crashes in the future if someone implements `operatorName()` differently.

cataphract (approved):

PR #12, opened 2021-10-05
Ruleset v2.1 Support

Anilm3:


cbeauchesne:
Good example of a PR that requires an update of the fuzzer. Please ping me if you need any help.

Anilm3:
> Good example of a PR that requires an update of the fuzzer. Please ping me if you need any help.

@cbeauchesne I have "fixed" the corpus generator, however I think it would benefit from a better `key_path` generation as well as `key_path` values.

fuzzing/scripts/build_corpus.py
```diff
@@ -265,7 +265,34 @@ def get_random_operation():
             return operation
 
         def get_random_address_array():
-            return _get_random_array2(self.addresses, 1, self.address_max_count, allow_none=False, unique=True)
+            addresses = _get_random_array2(self.addresses, 1, self.address_max_count, allow_none=False, unique=True)
```
cbeauchesne, line 268: I'm not sure I get the point of this logic ?

Side note, should it be at the begining of `get_payload()`? We geenrate `address_max_count` adresses, and always use those adresses.

cbeauchesne: I've added a small commit for namings

src/PWManifest.cpp
```diff
@@ -63,6 +43,14 @@ const PWManifest::ArgDetails& PWManifest::getDetailsForTarget(const PWManifest::
     return argManifest.find(argID)->second;
 }
 
+PWManifest::ArgDetails& PWManifest::getDetailsForTarget(const std::string& target)
+{
+    // We can't really return a dummy object when the key doesn't exist so the caller need to call `hasTarget` first.
```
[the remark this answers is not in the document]
Anilm3: Incidentally I don't need this particular method anymore.

cbeauchesne (approved):

src/PWManifest.cpp
```diff
@@ -7,44 +7,24 @@
...
     argManifest.reserve(count);
 }
 
-void PWManifest::insert(std::string_view name, PWManifest::ArgDetails&& arg)
+PWManifest::ARG_ID PWManifest::insert(std::string_view name, PWManifest::ArgDetails&& arg)
 {
-    argManifest.emplace(counter, std::move(arg));
+    auto [it, result] = argManifest.emplace(counter, std::move(arg));
+    (void) result; // unused
     argIDTable.emplace(name, counter);
```
cataphract, line 20: Not clear if you actually intend this to be a noop if `name` was added before. Perhaps you meant an error in this case. I also don't understand why `argManifest` is a map of sequential ints as keys rather than using a vector for this.

PR #21, opened 2021-10-21
Fix build if CMAKE_INSTALL_PREFIX is changed

cataphract:


Anilm3 (approved):
Seems harmless enough...

PR #26, opened 2021-11-10
ddwaf_result: add timeout flag and remove action field

Anilm3:
- Removal of `ddwaf_result.action` due to being equivalent to the return value of `ddwaf_run`.
- Removal of `DDWAF_ERR_TIMEOUT`.
- Addition of `ddwaf_result.timeout` to signal when there has been a timeout at any point during the run.
- Rearranged items in `ddwaf_result` to reduce struct size.

In case of timeout, the WAF will return a verdict (`DDWAF_GOOD`, `DDWAF_BLOCK`, `DDWAF_MONITOR`) and set the flag if there has been a timeout during the operation regardless of when it happens.

### DIFF
--- a/src/PWAdditive.cpp
+++ b/src/PWAdditive.cpp
@@ -30,7 +30,12 @@ PWAdditive::PWAdditive(const ddwaf_handle _waf, ddwaf_object_free_fn free_fn)
     argCache.reserve(ADDITIVE_BUFFER_PREALLOC);
 }
 
-ddwaf_result PWAdditive::run(ddwaf_object newParameters, uint64_t timeLeft)
+DDWAF_RET_CODE PWAdditive::run(ddwaf_object newParameters, uint64_t timeLeft) {
+    return run(newParameters, std::nullopt, timeLeft);
+}
+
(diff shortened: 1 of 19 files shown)


src/PWAdditive.cpp
```diff
@@ -50,7 +55,11 @@ ddwaf_result PWAdditive::run(ddwaf_object newParameters, uint64_t timeLeft)
     // consistent across all possible timeout scenarios.
     if (timeLeft == 0)
     {
-        return returnErrorCode(DDWAF_ERR_TIMEOUT);
+        if (res) {
+            ddwaf_result &output = res.value();
+            output.timeout = true;
```
cataphract, line 60: If all you need is an out-parameter for whether a timeout occurred, it would make more sense to pass a pointer/reference to a bool only, rather than the full `ddwaf_result`.

cataphract (approved):

src/PWAdditive.cpp
Anilm3: `ddwaf_result` is used below after the run.

cataphract (approved):

PR #31, opened 2021-11-24
Fix MacOS compilation errors for 10.9

nizox:
The MacOS toolchain has incomplete support for C++17 on 10.9 (which is the minimum version required for distributing Python binaries on MacOS). This PR builds the MacOS libraries for this target and workarounds the unsupported method.

### DIFF
--- a/src/PWAdditive.cpp
+++ b/src/PWAdditive.cpp
@@ -55,8 +55,8 @@ DDWAF_RET_CODE PWAdditive::run(ddwaf_object newParameters,
     // consistent across all possible timeout scenarios.
     if (timeLeft == 0)
     {
-        if (res) {
-            ddwaf_result &output = res.value();
+        if (res.has_value()) {
+            ddwaf_result &output = *res;
             output.timeout = true;
         }
         return DDWAF_GOOD;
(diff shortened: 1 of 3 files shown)


src/PWAdditive.cpp
```diff
@@ -55,8 +55,8 @@ DDWAF_RET_CODE PWAdditive::run(ddwaf_object newParameters,
     // consistent across all possible timeout scenarios.
     if (timeLeft == 0)
     {
-        if (res) {
-            ddwaf_result &output = res.value();
+        if (res.has_value()) {
```
Anilm3, line 58: If you set the minimum version but don't actually make these changes does it still build in CI? Because if that is the case we don't really need them.

Anilm3 (approved):
Approved, although if you find other issues with c++17 compatibility we should consider trying to figure out how to fix the linking issues instead.

src/PWAdditive.cpp
nizox: It doesn't and fails with the following message: 
```
    /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-req-build-885l7kqt/build/temp.macosx-10.9-x86_64-3.6/ddtrace.appsec._ddwaf_0/libddwaf-prefix/src/libddwaf/src/PWAdditive.cpp:84:36: error: 'value' is unavailable: introduced in macOS 10.14
            ddwaf_result &output = res.value();
```

Link to the full build logs:
https://github.com/DataDog/dd-trace-py/runs/4302682573?check_suite_focus=true#step:5:878

PR #42, opened 2022-01-04
Fix an integer overflow in the log macro

Taiki-San:
Fix an integer overflow in `DDWAF_LOG_HELPER`. This is likely unexploitable since `malloc(INT32_MIN)` should return `NULL` which causes an early abort.
Also solves a bunch of implicit casts warnings.

### DIFF
--- a/src/PWRetriever.cpp
+++ b/src/PWRetriever.cpp
@@ -78,7 +78,7 @@ bool PWRetriever::PWArgsWrapper::_validate_object(const ddwaf_object& input, uin
                 }
                 else
                 {
-                    DDWAF_TRACE("Performing recursive validation of item #%zu", i);
+                    DDWAF_TRACE("Performing recursive validation of item #%llu", i);
                 }
 
                 if (!_validate_object(array[i], depth + 1))
--- a/src/PWRetriever.cpp
+++ b/src/PWRetriever.cpp
@@ -78,7 +78,7 @@ bool PWRetriever::PWArgsWrapper::_validate_object(const ddwaf_object& input, uin
                 }
                 else
                 {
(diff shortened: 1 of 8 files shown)


src/PWRetriever.cpp
```diff
@@ -78,7 +78,7 @@ bool PWRetriever::PWArgsWrapper::_validate_object(const ddwaf_object& input, uin
                 }
                 else
                 {
-                    DDWAF_TRACE("Performing recursive validation of item #%zu", i);
+                    DDWAF_TRACE("Performing recursive validation of item #%llu", i);
```
cataphract, line 81: `PRIu64` is the portable macro to use, though in practice I don't think any compiler makes `sizeof(unsigned long long) == 16`

src/PowerWAFInterface.cpp
```diff
@@ -99,11 +99,12 @@ extern "C"
 
         PowerWAF* waf   = reinterpret_cast<PowerWAF*>(handle);
         auto& addresses = waf->manifest.get_root_addresses();
-        *size           = addresses.size();
-        if (addresses.empty())
+        if (addresses.empty() || addresses.size() > UINT32_MAX)
         {
             return nullptr;
```
cataphract, line 104: Should also set `*size` to `0`, to be safest and consistent with the first early return.
Also, this being C++, `numeric_limits` can be used, e.g. `std::numeric_limits<uint32_t>::max()` or `std::numeric_limits<std::remove_pointer_t<decltype(size)>>::max()`

cataphract (approved):

cataphract (approved):

PR #43, opened 2022-01-06
Introduce a script to run the WAF on a single rule & some test vectors

Taiki-San:


examples/verify_rule.cpp
```diff
@@ -0,0 +1,221 @@
...
+	bool success = true;
+	std::string ruleID = rule["id"].as<std::string>();
+	YAML::Node matches = rule["test_vectors"][runPositiveMatches ? "matches" : "no_matches"];
+	if (matches != nullptr)
+	{
+		size_t counter = 0;
+		for (YAML::const_iterator vector = matches.begin(); vector != matches.end(); ++vector, ++counter) {
+			ddwaf_object root = vector->as<ddwaf_object>();
+			if(root.type != DDWAF_OBJ_INVALID) {
+				ddwaf_context ctx = ddwaf_context_init(handle, NULL);
```
Anilm3, line 147: You can pass `ddwaf_object_free` as the second parameter and avoid the call to `ddwaf_object_free`.

Anilm3 (approved):
Looks fine to me, since it's an example/tool it should be fine.

examples/verify_rule.cpp
Taiki-San, line 147:
```suggestion
				ddwaf_context ctx = ddwaf_context_init(handle, ddwaf_object_free);
```

Anilm3 (approved):

PR #47, opened 2022-01-28
Add transformer to run on the keys of a parameter

Taiki-San:
Introduce two new transformers:
- `run_on_keys`: Run on the parameter keys
- `run_on_values`: Run on the parameter values

The two transformers can be combined, to run on both.
They're otherwise exclusive (none = `run_on_values`, either = not running on the other, both = run on both).

This is implemented by creating a virtual manifest entry with the correct `runOnKey`/`runOnValue` bit, thus enabling no runtime cost. To make the ARG_ID unique, we added a mask with `runOnKey << 1 | runOnValue` to efficiently find which entry to use.

### DIFF
--- a/src/PWManifest.cpp
+++ b/src/PWManifest.cpp
@@ -17,7 +17,7 @@ PWManifest::ARG_ID PWManifest::insert(std::string_view name, PWManifest::ArgDeta
 {
     auto [it, result] = argManifest.emplace(counter, std::move(arg));
     (void) result; // unused
-    argIDTable.emplace(name, counter);
+    argIDTable.emplace(std::pair<std::string, uint8_t>(name, RUN_ON_MASK(arg.runOnKey, arg.runOnValue)), counter);
 
     if (root_address_set.find(it->second.inheritFrom) == root_address_set.end())
     {
@@ -28,14 +28,14 @@ PWManifest::ARG_ID PWManifest::insert(std::string_view name, PWManifest::ArgDeta
     return counter++;
 }
 
-bool PWManifest::hasTarget(const std::string& string) const
+bool PWManifest::hasTarget(const std::string& string, uint8_t runOnMask) const
 {
-    return argIDTable.find(string) != argIDTable.cend();
+    return argIDTable.find(std::pair<std::string, uint8_t>(string, runOnMask)) != argIDTable.cend();
 }
 
-PWManifest::ARG_ID PWManifest::getTargetArgID(const std::string& target) const
+PWManifest::ARG_ID PWManifest::getTargetArgID(const std::string& target, uint8_t runOnMask) const
 {
-    return argIDTable.find(target)->second;
+    return argIDTable.find(std::pair<std::string, uint8_t>(target, runOnMask))->second;
 }
 
 const PWManifest::ArgDetails& PWManifest::getDetailsForTarget(const PWManifest::ARG_ID& argID) const
@@ -52,7 +52,7 @@ const std::string& PWManifest::getTargetName(const PWManifest::ARG_ID& target) c
     {
         if (argIDDefinition.second == target)
         {
-            return argIDDefinition.first;
+            return argIDDefinition.first.first;
         }
     }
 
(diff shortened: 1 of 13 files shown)


Anilm3:
I was looking into making some of the changes we discussed in this PR but these transformers as they are do not fit the semantics of the ruleset. 

For example, If you consider the fact that transformers apply one after another, if we were to implement in the future an actual transformer that converts a map into an array (for example), you can't have a `run_on_key` and then a `run_on_value` because there would be no values to run on. If anything you'd have a `run_on_key` or `run_on_key_and_value`.

What do you think?

Taiki-San:
I agree that those two transformers don't respect the standard semantic, in part because they're a bit a of a hack to address the manifest without making it explicitly available in the format.
We could address your exemple by mentioning in the documentation that those transformers must be at the end of the list to make the semantic a bit less ambiguous. This doesn't have to be true but would make reading the list of transformers make more sense. That would be helped by introducing an explicit `run_on_key_and_value`, since we would never need more than one such transformer at the end of the transformer list.

Anilm3:
> I agree that those two transformers don't respect the standard semantic, in part because they're a bit a of a hack to address the manifest without making it explicitly available in the format. We could address your exemple by mentioning in the documentation that those transformers must be at the end of the list to make the semantic a bit less ambiguous. This doesn't have to be true but would make reading the list of transformers make more sense. That would be helped by introducing an explicit `run_on_key_and_value`, since we would never need more than one such transformer at the end of the transformer list.

The new transformer should respect the semantics, i.e. once used there are no values to match (this doesn't apply to reporting). Documented or undocumented exceptions will inevitably be misused.

What I recommend to get this PR forward is to remove the `run_on_values` transformer and perhaps simplify the way the manifest entry is created, for example you can append something to the string (`@keys`?) and avoid having a separate parameter.

Taiki-San:
> The new transformer should respect the semantics, i.e. once used there are no values to match (this doesn't apply to reporting). Documented or undocumented exceptions will inevitably be misused.

Agree with you on that, but the solution then is to reject this transformer if it's not in the last position, no? Ignoring keys after some arbitrary point in transformer processing sounds quite annoying for absolutely no benefit :(

> What I recommend to get this PR forward is to remove the `run_on_values` transformer

We have a bunch of rules looking to run on both keys and values. They won't be part of the first wave but it would be nice to have a way to run on both keys and values. I'm up to replace `run_on_values` by `run_on_key_and_value` if you like it better.

> simplify the way the manifest entry is created, for example you can append something to the string (@keys?) and avoid having a separate parameter, although this will look strange on the logs.

I'm generally very afraid of leaking internal context in the public API: such a change would strictly forbid the use of magic character sequences in adresses (otherwise, we may have a collision resulting in a lot of tears). It'll also make reporting the matched address a bit harder (since the library/backend doesn't know about `@keys`, we'll need to strip it).
  
If the current approach doesn't work for you, I'd rather explore another option (bit mask in the `ARG_ID`?).

Anilm3:
> Agree with you on that, but the solution then is to reject this transformer if it's not in the last position, no? Ignoring keys after some arbitrary point in transformer processing sounds quite annoying for absolutely no benefit :(

As discussed, in an ideal WAF model, this transformer would convert the input object into an array of keys, so you would expect subsequent transformers to apply to the keys rather than the values. However at the moment this is not the case and it might not ever be, so in practice this transformer prevents any other transformer from being applied, regardless of the order.

> We have a bunch of rules looking to run on both keys and values. They won't be part of the first wave but it would be nice to have a way to run on both keys and values. I'm up to replace run_on_values by run_on_key_and_value if you like it better.

I think it's best if we don't get ahead of ourselves when it comes to adding new features which involve both the WAF input and output, as there will be RFCs in which this will be discussed. For example, the `run_on_key` transformer is already requiring a relatively significant change to the output. So let's just keep `run_on_key` for now.

> I'm generally very afraid of leaking internal context in the public API: such a change would strictly forbid the use of magic character sequences in adresses (otherwise, we may have a collision resulting in a lot of tears). It'll also make reporting the matched address a bit harder (since the library/backend doesn't know about @keys, we'll need to strip it).

That's fair enough, but please provide appropriate constants for the current magic numbers.

Anilm3 (approved):

PR #55, opened 2022-03-02
Ruleset parsing diagnostics

Anilm3:
This PR introduces `ddwaf_ruleset_info`, parsing diagnostics provided through `ddwaf_init`. These include:
- Failed rules
- Loaded rules
- Errors for failed rules when possible, i.e. the ID was available
- Version of the ruleset

### DIFF
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -77,7 +77,7 @@ typedef struct _ddwaf_object ddwaf_object;
 typedef struct _ddwaf_config ddwaf_config;
 typedef struct _ddwaf_result ddwaf_result;
 typedef struct _ddwaf_version ddwaf_version;
-
+typedef struct _ddwaf_ruleset_info ddwaf_ruleset_info;
 /**
  * @struct ddwaf_object
  *
@@ -93,7 +93,7 @@ struct _ddwaf_object
         const char* stringValue;
         uint64_t uintValue;
         int64_t intValue;
-        const ddwaf_object* array;
+        ddwaf_object* array;
     };
     uint64_t nbEntries;
     DDWAF_OBJ_TYPE type;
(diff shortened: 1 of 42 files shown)


src/ruleset_info.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+
+void ruleset_info::insert_error(std::string_view rule_id, std::string_view error)
+{
+    if (info == nullptr)
+    {
+        return;
+    }
+
+    ddwaf_object *rules, id, tmp;
```
cataphract, line 32: `tmp` need not be defined at this scope.

src/ruleset_info.cpp
```diff
@@ -0,0 +1,61 @@
...
+    else
+    {
+        rules = it->second;
+    }
+
+    ddwaf_object_stringl(&id, rule_id.data(), rule_id.size());
+    ddwaf_object_array_add(rules, &id);
+
+    add_failed();
+}
```
cataphract, line 59: The naming could be improved here a bit. There are variables similarly named (`errors` — which should probably be named sth like `error_ddwaf_obj_cache` — `error`, `info->errors`, `it`, `id`, `rule_id`). The types also don't help much because they're not contextually available or in the case of ddwaf_object you need to look at usage to determine the type of ddwaf_object.

include/ddwaf.h
```diff
@@ -143,6 +143,23 @@ struct _ddwaf_version
...
+ *
+ * Structure containing diagnostics on the provided ruleset.
+ * */
+struct _ddwaf_ruleset_info
+{
+    /** Number of rules successfully loaded **/
+    uint16_t loaded;
+    /** Number of rules which failed to parse **/
+    uint16_t failed;
+    /** Map of rule parsing errors {error: [rules]} **/
```
cataphract, line 157: I think describing it as _a map from an error string to an array of all the rule ids for which that error was raised_ would be clearer.

cataphract (approved):

PR #56, opened 2022-03-04
Metrics collector

Anilm3:


src/PWProcessor.hpp
```diff
@@ -13,29 +13,33 @@
...
-    ddwaf::condition::status hasCacheHit(const std::string& ruleID) const;
+    ddwaf::condition::status hasCacheHit(ddwaf::rule::index_type rule_idx) const;
     bool shouldIgnoreCacheHit(const std::vector<ddwaf::condition>& rules) const;
 
 public:
-    PWProcessor(PWRetriever& input, const ddwaf::rule_map& rules);
-    void startNewRun(const SQPowerWAF::monotonic_clock::time_point& _deadline);
-    void runFlow(const std::string& name, const std::vector<std::string>& flow, PWRetManager& manager);
+    PWProcessor(PWRetriever& input, const ddwaf::rule_vector& rules);
+    void startNewRun(const ddwaf::monotonic_clock::time_point& _deadline);
```
cataphract, line 38: this passing the timepoint (a plain 64bit int) by reference is probably a pessimization because it will require dereference of the pointer in rsi.

src/PowerWAFInterface.cpp
```diff
@@ -148,8 +150,21 @@ extern "C"
         }
         try
         {
+            // TODO: make the collector optional
+            optional_ref<ddwaf::metrics_collector> mc { std::nullopt };
+            if (collector != nullptr)
+            {
+                mc = *reinterpret_cast<ddwaf::metrics_collector*>(collector);
```
cataphract, line 157: this probably breaks strict aliasing rules, and could be avoided by having `ifdef __cplusplus` in the headers to make `ddwaf_metrics_collector = ddwaf::metrics_collector*`. Another option is to use the may_alias attribute.

src/PowerWAFInterface.cpp
```diff
@@ -173,9 +188,7 @@ extern "C"
 
         try
         {
-            PWAdditive* additive = reinterpret_cast<PWAdditive*>(context);
-            additive->flushCaches();
-            delete additive;
+            delete reinterpret_cast<PWAdditive*>(context);
```
cataphract, line 191: idem

src/metrics.cpp
```diff
@@ -0,0 +1,75 @@
...
+        {
+            continue;
+        }
+
+        // TODO this is weird
+        const std::string& rule_id = rules_[i].id;
+        ddwaf_object duration;
+        ddwaf_object_unsigned_force(&duration, rule_runtime_[i]);
+        ddwaf_object_map_addl(&final_metrics.rule_runtime,
+                              rule_id.c_str(), rule_id.size(), &duration);
```
cataphract, line 30: I don't like that this will do allocations and copy every single rule name. We're talking 100+ allocations+copy that will have to be done on every request. Also the final size of the map is known in advance, but the ddwaf_objects are in the map are not preallocated.

include/ddwaf.h
```diff
@@ -161,6 +157,19 @@ struct _ddwaf_ruleset_info
...
 
+/**
+ * @ddwaf_metrics
+ *
+ * Structure containing total and per-rule runtimes
+ **/
+struct _ddwaf_metrics
+{
+    /** Total WAF runtime in microseconds **/
+    uint64_t total_runtime;
```
cataphract, line 168: Why not stick to one unit? 2^64-1 ns is like 585 years, so it's not like we need to reduce precision to get magnitude.

Anilm3: Fair, I was also thinking about changing it.

src/PowerWAFInterface.cpp
Anilm3: The former might be okay, but the latter doesn't seem to have an alternative on MSVC.

src/PWProcessor.hpp
Anilm3: `time_point` is a class.

src/metrics.cpp
Anilm3: Actually the size of the map is not known beforehand as we only include those with non-zero duration.

src/PowerWAFInterface.cpp
Anilm3: I've changed it to use an ifdef, it's not pretty but it avoids all the reinterpret_cast shenanigans.

cataphract (approved):

PR #57, opened 2022-03-07
DDWAF Object getters

Anilm3:


src/object.cpp
```diff
@@ -337,4 +338,89 @@ extern "C"
...
+    }
+
+    const char* ddwaf_object_get_key(ddwaf_object* object, size_t* length)
+    {
+        if (object == NULL || object->parameterName == NULL)
+        {
+            return NULL;
+        }
+
+        if (length)
```
Taiki-San, line 374: Should the length parameter be optional?
We have no guarantee that strings are NULL-terminated, do we?

Anilm3: We absolutely do have a guarantee for `ddwaf_object` returned by the WAF. The same is not necessarily applicable to input objects, depending on how they are created, but this API is not meant for that purpose.

include/ddwaf.h
```diff
@@ -477,6 +477,101 @@ bool ddwaf_object_map_addl(ddwaf_object *map, const char *key, size_t length, dd
...
+/**
+ * ddwaf_object_type
+ *
+ * Returns the type of the object.
+ *
+ * @param object The object from which to get the type.
+ *
+ * @return The object type of DDWAF_OBJ_INVALID if NULL.
+ **/
+DDWAF_OBJ_TYPE ddwaf_object_type(ddwaf_object *object);
```
cataphract, line 489: I'd have preferred definitions with `inline`. You could then add a .c file compiled with `-std=c11` where you just had `extern inline DDWAF_OBJ_TYPE ddwaf_object_type();`. This should work for both C and C++ users of the header.

cataphract (approved):

PR #66, opened 2022-04-03
WAF Obfuscator

Anilm3:
WAF Obfuscator provided as a separate class, for now this simply provides a verdict on whether obfuscation should be done, but the objective would be to have an intermediate event structure which can then be directly obfuscated instead. As of now this introduces an inefficiency in that matches can be discarded after they have been created, if other conditions do not match, so the obfuscation would've been unnecessary (the same applies to generating JSON at that stage). This inefficiency will be addressed with the intermediate event structure.

This also introduces a slight refactoring of the object limits, for now the `max_string_length` limit is not being used but it will be in the near future.

### DIFF
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -17,8 +17,8 @@ extern "C"
 #include <stddef.h>
 
 #define DDWAF_MAX_STRING_LENGTH 4096
-#define DDWAF_MAX_MAP_DEPTH 20
-#define DDWAF_MAX_ARRAY_SIZE 256
+#define DDWAF_MAX_CONTAINER_DEPTH 20
+#define DDWAF_MAX_CONTAINER_SIZE 256
 #define DDWAF_RUN_TIMEOUT 5000
 
 /**
@@ -115,14 +115,15 @@ struct _ddwaf_object
 struct _ddwaf_config
 {
     struct {
-        /** Maximum size of ddwaf::object arrays and maps. */
-        uint32_t max_array_size;
-        /** Maximum depth of ddwaf::object maps. */
-        uint32_t max_map_depth;
+        /** Maximum size of ddwaf::object containers. */
+        uint32_t max_container_size;
+        /** Maximum depth of ddwaf::object containers. */
+        uint32_t max_container_depth;
         /** Maximum length of ddwaf::object strings. */
         uint32_t max_string_length;
     } limits;
 
+    /** Obfuscator regexes - the strings are owned by the caller */
     struct {
         /** Regular expression for key-based obfuscation */
         const char *key_regex;
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -18,7 +18,7 @@ extern "C"
 
 #define DDWAF_MAX_STRING_LENGTH 4096
 #define DDWAF_MAX_MAP_DEPTH 20
-#define DDWAF_MAX_ARRAY_LENGTH 256
+#define DDWAF_MAX_ARRAY_SIZE 256
 #define DDWAF_RUN_TIMEOUT 5000
 
 /**
@@ -114,10 +114,21 @@ struct _ddwaf_object
  **/
 struct _ddwaf_config
 {
-    /** Maximum length of ddwaf::object arrays. */
-    uint64_t maxArrayLength;
-    /** Maximum depth of ddwaf::object maps. */
-    uint64_t maxMapDepth;
+    struct {
+        /** Maximum size of ddwaf::object arrays and maps. */
+        uint32_t max_array_size;
+        /** Maximum depth of ddwaf::object maps. */
+        uint32_t max_map_depth;
+        /** Maximum length of ddwaf::object strings. */
+        uint32_t max_string_length;
+    } limits;
+
+    struct {
+        /** Regular expression for key-based obfuscation */
+        const char *key_regex;
+        /** Regular expression for value-based obfuscation */
+        const char *value_regex;
+    } obfuscator;
 };
 
 /**
--- a/src/PWRet.cpp
+++ b/src/PWRet.cpp
@@ -75,7 +75,7 @@ void PWRetManager::recordRuleMatch(const std::unique_ptr<IPWRuleProcessor>& proc
                 continue;
             }
 
-            redact = redact || event_obfuscator.obfuscate_key(
+            redact = redact || event_obfuscator.is_sensitive_key(
                 {key.stringValue, static_cast<size_t>(key.nbEntries)});
 
             jsonKey.SetString(key.stringValue, static_cast<rapidjson::SizeType>(key.nbEntries), allocator);
@@ -92,8 +92,8 @@ void PWRetManager::recordRuleMatch(const std::unique_ptr<IPWRuleProcessor>& proc
     highlight.SetArray();
 
     redact = redact ||
-             event_obfuscator.obfuscate_value(gather.resolvedValue) ||
-             event_obfuscator.obfuscate_value(gather.matchedValue);
+             event_obfuscator.is_sensitive_value(gather.resolvedValue) ||
+             event_obfuscator.is_sensitive_value(gather.matchedValue);
 
     if (redact) {
         param.AddMember("value", redaction_msg, allocator);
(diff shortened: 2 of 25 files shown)


include/ddwaf.h
```diff
@@ -114,10 +114,22 @@ struct _ddwaf_object
...
+    struct {
+        /** Maximum size of ddwaf::object containers. */
+        uint32_t max_container_size;
+        /** Maximum depth of ddwaf::object containers. */
+        uint32_t max_container_depth;
+        /** Maximum length of ddwaf::object strings. */
+        uint32_t max_string_length;
+    } limits;
+
+    /** Obfuscator regexes - the strings are owned by the caller */
```
Julio-Guerra, line 126: What about a simpler API by copying them instead and releasing them on ddwaf_destroy?

src/obfuscator.cpp
```diff
@@ -0,0 +1,64 @@
...
+        key_regex = std::make_unique<re2::RE2>(sp, options);
+    }
+
+    if (!value_regex_str.empty()) {
+        const re2::StringPiece sp(value_regex_str.data(), value_regex_str.size());
+        value_regex = std::make_unique<re2::RE2>(sp, options);
+    }
+}
+
+bool obfuscator::obfuscate_key(std::string_view key) const
```
Julio-Guerra, line 45: I'd rename this into `is_sensitive_key` or similar.

src/PWRet.cpp
```diff
@@ -69,14 +87,31 @@ void PWRetManager::recordRuleMatch(const std::unique_ptr<IPWRuleProcessor>& proc
...
+             event_obfuscator.obfuscate_value(gather.resolvedValue) ||
+             event_obfuscator.obfuscate_value(gather.matchedValue);
+
+    if (redact) {
+        param.AddMember("value", redaction_msg, allocator);
+        if (!gather.matchedValue.empty())
+        {
+            matchedValue.SetString(redaction_msg, allocator);
+            highlight.PushBack(matchedValue, allocator);
+        }
```
Julio-Guerra, line 104: This is right only if the key_path is sensitive. Otherwise, you're supposed to only obfuscate the substrings matching the value regexp.

Pseudo example:
`password: anything` becomes `password: <redacted by datadog>`
while `key: i am a SENSITIVE value` would become `key: i am a <redacted by datadog> value` where `key` is not considered sensitive by the key regexp and the value regexp matches `SENSITIVE`

include/ddwaf.h
```diff
@@ -114,10 +114,22 @@ struct _ddwaf_object
...
-    uint64_t maxArrayLength;
-    /** Maximum depth of ddwaf::object maps. */
-    uint64_t maxMapDepth;
+    struct {
+        /** Maximum size of ddwaf::object containers. */
+        uint32_t max_container_size;
+        /** Maximum depth of ddwaf::object containers. */
+        uint32_t max_container_depth;
+        /** Maximum length of ddwaf::object strings. */
+        uint32_t max_string_length;
```
Julio-Guerra, line 123: Note that changing those types now might make the upgrade work a bit harder.

src/PowerWAF.cpp
```diff
@@ -24,38 +24,75 @@
...
-      flows(std::move(flows_))
+namespace {
+obfuscator obfuscator_from_config(const ddwaf_config* config)
 {
-    if (config != nullptr)
-    {
-        if (config->maxArrayLength != 0)
+    std::string_view key_regex, value_regex;
+
+    if (config != nullptr) {
```
Julio-Guerra, line 32: Not enabling it by default is okay to me if you think everyone will be able to upgrade to this version easily.

src/obfuscator.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+    size_t length = findStringCutoff(value.data(), value.size());
+    const re2::StringPiece key_ref(value.data(), length);
+
+    return regex.Match(key_ref, 0, length, re2::RE2::UNANCHORED, nullptr, 0);
+}
+
+}
+
+obfuscator::obfuscator(std::string_view key_regex_str,
```
Julio-Guerra, line 26: What happens if the regexp cannot be compiled?
The lib should know so we can log an error.
I even wonder if we should start appsec in such as case.

Julio-Guerra (changes requested):

include/ddwaf.h
Anilm3: I don't want to copy the strings because I don't need them after the regex has been created. This allows you to initialise the structure with literals.

src/PWRet.cpp
Anilm3: You're right, I should've mentioned in the description that this was not implemented as the RFC. The main reason is that (as mentioned in the description) I plan to iterate on this to introduce two improvements:

- An intermediate event type, instead of generating JSON on matches.
- Obfuscation of the entire event, only when needed.

At this stage the obfuscator will actually obfuscate rather than provide a verdict and I should be able to be compliant with this.

Also, as we discussed, we won't be using value obfuscation yet so I introduced this opportunistically to avoid changing the config structure later.

include/ddwaf.h
Anilm3: I understand, but better one breaking change than many.

src/PowerWAF.cpp
Anilm3: We can't really do anything else if we want to provide flexibility, given that there are no defaults in C structs.

src/obfuscator.cpp
Anilm3: Literally nothing happens at the moment because RE2 fails silently. We can certainly log an error.

Anilm3: I have now added an error log if the regex fails to compile.

src/PWRet.cpp
Anilm3: If you have no objections to this, I can open an issue so that it's clear that it'll change in the near future.

include/ddwaf.h
Julio-Guerra: SGTM, as long as you are aware of this.

src/PWRet.cpp
Julio-Guerra: SGTM as a first step, especially because we know the value regexp wasn't required so far for the incidents we had 👍
So feel free to remove the value regexp for now if you want to, as part of this first iteration.

src/obfuscator.cpp
Anilm3: As discussed, I've changed it so that it'll fallback to the default key regex if it fails to compile the user-provided one and if that one fails to compile it'll fail to initialise the WAF.

src/PWRet.cpp
Anilm3: I'll keep it, but I'll make sure people are aware not to implement it for now.

src/obfuscator.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+class obfuscator
+{
+public:
+    explicit obfuscator(std::string_view key_regex_str = std::string_view(),
+        std::string_view value_regex_str = std::string_view());
+    bool obfuscate_key(std::string_view key) const;
+    bool obfuscate_value(std::string_view value) const;
+
+    static constexpr std::string_view redaction_msg{"<redacted by datadog>"};
```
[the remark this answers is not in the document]
Anilm3: Confirmed with @kit-cat that the correct string should be `<Redacted>`.

src/PWRet.hpp
```diff
@@ -17,19 +17,13 @@ class PWRetManager;
...
-    rapidjson::Document outputDocument;
-    rapidjson::Document::AllocatorType& allocator;
-    rapidjson::Value ruleCollector;
-
-    DDWAF_RET_CODE worstCode { DDWAF_GOOD };
-    bool timeout { false };
-
 public:
-    PWRetManager(rapidjson::Document::AllocatorType& allocator);
+    PWRetManager(const ddwaf::obfuscator &eo = ddwaf::obfuscator());
```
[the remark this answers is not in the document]
Anilm3: Yeah, you're right this is a dangling reference, it's only used for testing laziness but since it's UB I'll change it.

Julio-Guerra (approved):
The overall behaviour is good to me.
Note that I was unable to review the correctness of the C++ (😢)

PR #68, opened 2022-04-14
Correct nuget url

robertpi:


Anilm3 (approved):

Anilm3 (changes requested):
Actually I hadn't seen it but could you remove the `.vscode `stuff?

PR #75, opened 2022-05-18
Add per libc linux builds

lloeki:
Adds support for building libddwaf for multiple platforms and against different `libc`s in a controlled environment, whether run in CI, on Linux, or on macOS, and whether on an ARM or Intel machine.

This makes it seem that there are a lot of combinations, but each one of them proves to be useful in different scenarios:

- builds native to the targeted libc (glibc built on Debian, musl built on Alpine)
- symmetric, uniform experience whether the user is on an ARM (e.g Apple Silicon) or Intel machine
- fast cross-compiling to foreign platform
- ability to build as if on a foreign platform (albeit slow, via qemu)

Not all of them have to be run in CI, only those that aim to produce publishable artifacts. Nonetheless, every case has proven to be useful in practice, and this persists self-documenting, reproducible tooling that I previously had to piece together over and over again.

Also, updated `docker/README.md` with valuable information (see `Per-libc linux builds`).


Anilm3:
@lloeki given that there wasn't really an issue with the existing builds I think it's best if we don't muddy the water with more unnecessary combinations unless we can prove that the universal builds aren't universal.

lloeki:
@Anilm3 this goes beyond providing multiple additional builds but also covers support for working from M1 machines:

> symmetric, uniform experience whether the user is on an ARM (e.g Apple Silicon) or Intel machine
> fast cross-compiling to foreign platform
> ability to build as if on a foreign platform (albeit slow, via qemu)

The first one happens to be incidental:

> builds native to the targeted libc (glibc built on Debian, musl built on Alpine)

lloeki:
Closing this one in favor of #93

PR #78, opened 2022-05-26
Simplify manifest

Anilm3:
Simplify the use of the manifest:
- Reduce string operations.
- Automatically determine the parent / child relationship between addresses.
- Add support for `key_path.size() > 1`.
- Introduce manifest builder to avoid storing unnecessary information.


src/manifest.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+protected:
+    std::unordered_map<std::string, target_type> targets_{};
+    std::unordered_map<target_type, target_info> info_{};
+
+    // Root address memory to be returned to the API caller
+    std::vector<const char*> root_addresses_;
+};
+
+class manifest_builder {
```
estringana, line 96: Out of interest. Is it ok to have multiple classes per hpp file on C++?

estringana (approved):

src/manifest.hpp
Anilm3: Yes, although ideally they should be related somehow.

cataphract (approved):

src/manifest.hpp
```diff
@@ -0,0 +1,120 @@
...
+            hash ^= std::hash<std::string>{}(str);
+        }
+        return hash;
+    }
+};
+}
+
+namespace ddwaf
+{
+class manifest {
```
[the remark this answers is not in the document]
Anilm3: My objective was to remove it altogether but it's fine as a simple abstraction for now.

Anilm3:
I'll address any comments on a future PR since none of them are critical.

PR #79, opened 2022-06-01
Expose RE2

robertpi:
Make RE2 easier to user from languages without an RE2 implementation.

Experimentally for testing with .NET


Anilm3:
Exposing internal third-party libraries used by the WAF is not a good idea and it's also terrible in terms of library design, this is a WAF library after all. If you need a fully fledged RE2 wrapper then it's best if you create a separate wrapper for your purpose. If you just want to use the obfuscator, we can always make it a standalone object with its own API.

PR #80, opened 2022-06-09
Make libinjection look for backticks

Taiki-San:
Make libinjection more tolerant of backticks, since they're often used in SQL injections.
For instance, introduce detection for 
```
1` or "1" = "1"
```


Anilm3 (approved):
lgtm

PR #85, opened 2022-07-15
Remove validator

Anilm3:


estringana (approved):

src/manifest.hpp
```diff
@@ -60,12 +60,12 @@ class manifest {
...
-    target_type get_target(const std::string& name) const {
+    std::pair<bool, target_type> get_target(const std::string& name) const {
         auto it = targets_.find(name);
         if (it == targets_.end()) {
-            return {};
+            return {false, 0};
         }
-        return it->second;
+        return {true, it->second};
     }
```
cataphract, line 69: `std::optional` seems like a more explicit option here

cataphract (approved):

src/manifest.hpp
Anilm3: This is actually the approach used by the STL (e.g. emplace), albeit I think I flipped the order, but you're right that optional could be a good option as well. I'll think about it and consider updating it in the next PR.

PR #93, opened 2022-08-03
Improve build system

lloeki:
Derived from and obsoletes #75.

Adds support for building `libddwaf` for multiple platforms in a controlled Docker environment, whether run in CI, on Linux, or on macOS, and whether on an ARM or Intel machine:

- symmetric, uniform experience whether the user is on an ARM (e.g Apple Silicon) or Intel machine
- fast cross-compiling to foreign platform
- ability to build as if on a foreign platform (albeit slow, via qemu)

Also clarifies, simplifies, and unifies some build system and CI elements:
- CI job names are simpler and clearer, referring not to the build host but the build target
- macOS jobs are unified using a matrix, merely skipping tests for `arm64-darwin`

Also, updated `docker/README.md` with usage information, architecture documentation, and improving the description of how we produce portable builds.


Anilm3:
Feel free to update the branch and I can review it.

.github/workflows/build.yml
```diff
@@ -40,22 +45,39 @@ jobs:
...
+    name: macOS (${{ matrix.platform }})
     steps:
       - uses: actions/checkout@v2
         with:
           submodules: recursive
       - name: Create Build Directory
         run: cmake -E make_directory ${{ github.workspace }}/build ${{ github.workspace }}/packages
       - name: Generating Build Scripts
-        run: cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMACOSX_DEPLOYMENT_TARGET=10.9 -DCMAKE_INSTALL_PREFIX=${{ github.workspace }} -DCPACK_PACKAGE_DIRECTORY=${{ github.workspace }}/packages ${{ github.workspace }}
+        run: cmake -DCMAKE_OSX_ARCHITECTURES=${{ matrix.arch }} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMACOSX_DEPLOYMENT_TARGET=10.9 -DCMAKE_INSTALL_PREFIX=${{ github.workspace }} -DCPACK_PACKAGE_DIRECTORY=${{ github.workspace }}/packages ${{ github.workspace }}
```
cataphract, line 70: One thing I noticed with the current builds is that actually do not target 10.9, but rather 11.6:

```
$ curl -sL https://github.com/DataDog/libddwaf/releases/download/1.5.0/libddwaf-1.5.0-darwin-x86_64.tar.gz | tar -O - -xzf - libddwaf-1.5.0-darwin-x86_64/lib/libddwaf.dylib > /tmp/libddwaf.dylib; otool -l /tmp/libddwaf.dylib | grep minos
    minos 11.6
```

lloeki: IIRC I removed it either because of cross-compilation to arm64-darwin, which can't target below macOS 11 since that's the first version with `arm64-darwin` support or because the macOS CI runners were brownout'd and the updated ones refused to proceed with targeting macOS 10, but I can't recall which it was.

We can either make it a matrix variable if targeting 10.9 works for Intel (I'll try that), or drop support for macOS 10.

If the former works, I wonder if this value should be part of cmake instead of the CI config.

lloeki: Oh, wait, `-DMACOSX_DEPLOYMENT_TARGET=10.9` is still there. That might come from `runs-on: macos-latest` then?

lloeki:
@Anilm3 getting my head out to do the 1.5 update to the Ruby bindings, this just got rebased on tag `1.5.1`.

I recall having a few improvements not pushed yet, I just have to find them again in some git stash and finish them up.

lloeki:
Rebased

Anilm3:
@lloeki this is no longer required due to https://github.com/DataDog/libddwaf/pull/198 so I'll close the PR, unless you have any objections.

lloeki:
No objections, this is exactly where I wanted things to land!

Closing.

PR #94, opened 2022-08-17
Disable the 1)c libinjection fingerprint

Taiki-San:
libinjection work by computing a fingerprint from an input and matching it in a deny list database.
We disable the `1)c` fingerprint because it only triggered false positives from our analysis.


Anilm3 (approved):
LGTM

PR #99, opened 2022-08-31
Toggle rules at runtime

Anilm3:


src/rule.hpp
```diff
@@ -103,8 +105,31 @@ class rule
     rule(const rule&) = delete;
     rule& operator=(const rule&) = delete;
 
-    rule(rule&&) = default;
-    rule& operator=(rule&&) = default;
+    // Constructors are not atomic...
```
cataphract, line 108: I think the comment is a bit misleading. The problem is that the atomic is not movable.

cataphract (approved):

src/rule.hpp
Anilm3: Okay let me rephrase it to make clear that when using the move constructor there should be no expectation of the atomic behaving as such.

PR #102, opened 2022-09-06
GTest ddwaf_result validators

Anilm3:
- Add `WithEvents` to be used with `EXPECT_THAT(ddwaf_result, ...)`, this removes the need for raw JSON strings.
- Use `C++20` on tests (for designated initialisers).
- Update tests which do not currently require a rewrite.
- Validate event schema.
- Add `EXPECT_EVENT` to simplify the assertion of events (validate schema + validate event).


tests/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ add_executable(testPowerWAF ${LIBDDWAF_SOURCE} ${LIBDDWAF_TEST_SOURCE})
 option(LIBDDWAF_TEST_COVERAGE "Do coverage if possible" OFF)
 
 set_target_properties(testPowerWAF PROPERTIES
-    CXX_STANDARD 17
+    CXX_STANDARD 20
```
estringana, line 9: Out of curiosity since I have no much experience with c++. Is it that easy to change from on standard to another? I mean it does not have any collateral impact anywhere else or having to change code that was on 17?

tests/event_serializer_test.cpp
```diff
@@ -57,19 +57,25 @@ TEST(TestEventSerializer, SerializeSingleEventSingleMatch)
...
-    std::unordered_set<std::string_view> expected_actions{"block", "monitor"};
-    std::unordered_set<std::string_view> found_actions;
-    for (unsigned i = 0; i < output.actions.size; i++) {
-        char *value = output.actions.array[i];
-        EXPECT_NE(value, nullptr);
-        found_actions.emplace(value);
-    }
-
-    EXPECT_EQ(expected_actions, found_actions);
+    EXPECT_EVENT(output,
```
estringana, line 61: This is quite cool! Bye Json

estringana (approved):

tests/CMakeLists.txt
Anilm3: Generally, language changes to the standard are backwards compatible, STL changes on the other hand aren't as they might deprecate older methods and functions in favour of new ones. Having new classes, methods and functions can potentially cause issues if dynamically linking against a version of libstdc++ or libc++ missing these new symbols. However in this case the standard change is limited to the tests.

PR #107, opened 2022-09-23
Strip libddwaf.a for darwin/linux release

Hellzy:
Add strip invocation on libddwaf.a for linux and darwin in release builds.
libddwaf.a is used by the Go tracer to link with the WAF and stripping makes for a better user experience.
This emerged from https://github.com/DataDog/dd-trace-go/issues/1478.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,7 +208,7 @@ if (LIBDDWAF_BUILD_STATIC)
               endif()
               add_custom_command(TARGET libddwaf_static POST_BUILD
                   COMMAND ${STRIP} -x -S $<TARGET_FILE:libddwaf_static> -o $<TARGET_FILE:libddwaf_static>.stripped)
-              install(FILES $<TARGET_FILE:libddwaf_static>.stripped DESTINATION${CMAKE_INSTALL_LIBDIR})
+              install(FILES $<TARGET_FILE:libddwaf_static>.stripped DESTINATION ${CMAKE_INSTALL_LIBDIR})
           endif()
       endif()
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,8 +198,18 @@ if (LIBDDWAF_BUILD_STATIC)
               COMMAND ${CMAKE_COMMAND} -E copy ar_comb/combined${CMAKE_STATIC_LIBRARY_SUFFIX} $<TARGET_FILE:libddwaf_static>
               COMMAND rm -rf ar_comb
               WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
-        endif()
-    endif()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -260,6 +260,8 @@ if (LIBDDWAF_BUILD_SHARED)
             else()
                 add_custom_command(TARGET libddwaf_shared POST_BUILD
                     COMMAND ${STRIP} -S -x $<TARGET_FILE:libddwaf_shared>)
+                add_custom_command(TARGET libddwaf_static POST_BUILD
```
Anilm3, line 263: There's no guarantee that the static library will be built here, notice that this is inside an if statement `if (LIBDDWAF_BUILD_SHARED)`.

This should be done above inside https://github.com/DataDog/libddwaf/pull/107/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR149

Also the debug symbols should be stored as well.

CMakeLists.txt
```diff
@@ -271,6 +273,8 @@ if (LIBDDWAF_BUILD_SHARED)
             else()
                 add_custom_command(TARGET libddwaf_shared POST_BUILD
                     COMMAND ${CMAKE_STRIP} $<TARGET_FILE:libddwaf_shared>)
+                add_custom_command(TARGET libddwaf_static POST_BUILD
```
Anilm3, line 276: Idem

CMakeLists.txt
Hellzy: Storing debug symbols for static libraries on MacOs seems to be quite a painful process.
I managed to do so by extracting all .o files from the unstripped .a and invoking `dsymutil` on those, however dsymutil seems to complain that there's no debug symbol to be found, which can be confirmed by calling `dsymutil -dump-debug-map` on the generated .dwarf files.
FYI here is how I configure cmake and build:
` cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -Wno-dev`
 `make -j8`
 
 @Anilm3 have you had any experience with this?

Anilm3: Unfortunately, I don't have any advice, I'm surprised that the error is that the symbols are missing. Perhaps @lloeki has advice as he has more experience with MacOs.

Alternatively, as a measure of last resort if you find that this can't be achieved, consider keeping both stripped and original versions of the `.a`.

Julio-Guerra:
FTR, this is needed for us before the next libddwaf release 🙏

Hellzy:
@Anilm3 this new version publishes a striped and an unstriped version. Also it doesn't take more space than publishing debug symbols on the side

Anilm3 (approved):

PR #111, opened 2022-11-08
Experiment building libddwaf on the oldest available macos target

Julio-Guerra:
We are currently compiling libddwaf on `macos-latest` where it appears some new symbols have been introduced in the libstdc++ which are absent for macos 10.15's libstdc++. 

Related to https://github.com/DataDog/dd-trace-go/issues/1478

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
           path: ${{ github.workspace }}/packages/*.tar.gz
 
   macos-build:
-    runs-on: macos-latest
+    runs-on: macos-10.15
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


Anilm3 (approved):
lgtm!

.github/workflows/build.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           path: ${{ github.workspace }}/packages/*.tar.gz
 
   macos-cross-build:
-    runs-on: macos-latest
+    runs-on: macos-10.15
```
Anilm3, line 74: @gnufede mentioned that macos support for arm64 started with big sur (11), so you can probably revert this one to `macos-latest`.

gnufede: You can also use `macos-11.0` if you want to have it pinned, IIRC

Julio-Guerra: I didn't check the details but it probably compiles because we cross-compile here.

Julio-Guerra: Pinned to macos-11 anyway 👍

PR #122, opened 2022-12-07
Rework path tree for exclusion

cataphract:


src/exclusion/object_filter.hpp
```diff
@@ -22,35 +24,131 @@
...
 
-    std::string_view get_stored_string(std::string_view str);
+public:
+    class traverser {
+        trie_node const *const cur_node{};
+    public:    
+        enum class state {
+            NOT_FOUND,
+            FOUND,
+            GLUE
```
Anilm3, line 90: nit: I don't usually use capital letters for enum class values since they are nicely namespaced and type-safe so it's already clear they are "constants".

Anilm3: Also I'm unconvinced about `GLUE`, it's unclear what that's supposed to mean, although I don't have good suggestions without referring to trie internals.

src/exclusion/object_filter.hpp
```diff
@@ -22,35 +24,131 @@
...
+    };
+
+    template <typename StringType,
+        typename = std::enable_if<std::is_constructible<std::string, StringType>::value>>
+    void insert(const std::vector<StringType> &path)
+    {
+        is_empty = false;
+
+        trie_node *cur = &root;
+        bool is_new;
```
Anilm3, line 116: If the path is empty this will be used uninitialised.

src/exclusion/object_filter.hpp
```diff
@@ -22,35 +24,131 @@
...
+            return children.empty();
+        }
     };
+    static_assert(std::is_move_assignable<trie_node>::value);
 
-    explicit path_trie(std::shared_ptr<trie_node> node_) : root(std::move(node_)) {}
+    trie_node root{""};
+    // we allow adding the root to the trie (matching everything)
+    // so we use this flag to distinguish the two cases (empty vs everything)
+    bool is_empty{true};
```
Anilm3, line 81: What do you think of making the root an `std::optional<trie_node>` and remove the `is_empty` variable altogether?

src/exclusion/object_filter.hpp
```diff
@@ -22,35 +24,131 @@
...
-
-protected:
+private:
     struct trie_node {
-        // Heterogenous lookups required...
-        std::unordered_map<std::string_view, std::shared_ptr<trie_node>> values;
-        // TODO this shouldn't be necessary
-        bool terminal{false};
+        std::string path_component;
+        std::vector<trie_node> children{};
```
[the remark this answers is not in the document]
cataphract: You mean with `struct trie_node { vector<pair<string, trie_node>> children; }`? I'm only seeing drawbacks: the pair impairs the legibility (the first template argument of the pair is not given a name), the iterator gets bigger because it needs to store some string reference for the path_component, and I probably I still need to provide a custom comparator.

src/exclusion/object_filter.hpp
```diff
@@ -22,35 +24,131 @@
...
+        struct comp {
+            bool operator()(const trie_node& n, std::string_view v) {
+                return std::string_view{n.path_component} < v;
+            }
+            bool operator()(std::string_view v, const trie_node& n) {
+                return v < std::string_view{n.path_component};
+            }
+        };
+
+        [[nodiscard]] trie_node const *get_child(std::string_view key) const {
```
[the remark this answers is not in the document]
cataphract: Most would probably benefit from the inlining. What about moving the whole thing inside object_filter.cpp? The only reason the type is public is because of testing, right?

src/exclusion/object_filter.hpp
cataphract: I'll rename it to something `INTERMEDIATE_NODE`

src/exclusion/object_filter.hpp
cataphract: yeah, better to use the pair...

src/exclusion/object_filter.hpp
cataphract: it's a good idea

src/exclusion/object_filter.hpp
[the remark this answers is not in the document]
cataphract: > As mentioned, I think the reduced cost of indirection will be dwarfed by the search cost.

I'll try to benchmark it.

src/exclusion/object_filter.hpp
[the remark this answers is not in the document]
Anilm3: Yes, moving it entirely to the cpp would make it difficult to test independently, on the other hand it's a bit of a balance between inlining and recompilation pain as when you modify a header it results in recompiling many files.

src/exclusion/object_filter.hpp
```diff
@@ -22,35 +26,145 @@
...
+            {
+                auto it = children.find(key);
+                if (it != children.end()) {
+                    return {it->second, false};
+                }
+            }
+
+            auto interned_str = std::forward<InternString>(intern_str_fun)(key);
+            auto [it, is_new] = children.emplace(std::piecewise_construct,
+                std::forward_as_tuple(interned_str), std::forward_as_tuple());
```
Anilm3, line 60: neat!

Anilm3 (approved):
Looks great now!

PR #123, opened 2022-12-08
Priority collections

Anilm3:
This PR introduces Priority Collections, a variation on the existing collection concept which favours maximising the numbers of actions obtained rather than obtaining matches.  The collections are evaluated before regular collections as well and are a response to the problem of having rules with side-effects which should be evaluated before rules without. 

Other changes in this PR:
- Reduced expected fuzzer coverage as it seems to be quite flaky at the moment.
- Started using run-clang-tidy for the parallel speedup.
- Many small changes to reduce the number of clang-tidy warnings.

Note that the validator tests will not be included here as the ruleset / tests have become unwieldy so the directory structure needs to be refactored to simplify writing new tests.


estringana (approved):

src/rule_processor/regex_match.cpp
```diff
@@ -4,18 +4,21 @@
...
 // Copyright 2021 Datadog, Inc.
 
+#include <array>
 #include <exception.hpp>
 #include <rule_processor/regex_match.hpp>
 
 namespace ddwaf::rule_processor {
 
-regex_match::regex_match(const std::string &regex_str, std::size_t minLength, bool caseSensitive)
+regex_match::regex_match(const std::string &regex_str, std::size_t minLength, bool case_sensitive)
```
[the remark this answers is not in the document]
Anilm3: Yes,  I've been slowly changing the coding style and conventions, but there are still many little things I've missed like this one.

src/ruleset.hpp
```diff
@@ -73,8 +79,10 @@ struct ruleset {
     std::unordered_map<std::string_view, exclusion::input_filter::ptr> input_filters;
     // Rules are ordered by ID
     std::unordered_map<std::string, rule::ptr> rules;
-    // Collections are ordered by rule.type
-    std::unordered_map<std::string, std::vector<rule::ptr>> collections;
+
+    // Collections
```
cataphract, line 83: It would be useful to continue documenting the meaning of the key of the map.

src/collection.hpp
```diff
@@ -0,0 +1,81 @@
...
+    collection &operator=(collection &&) = default;
+
+    virtual void insert(rule::ptr rule) { rules_.emplace_back(std::move(rule)); }
+
+    virtual void match(std::vector<event> &events,
+        std::unordered_set<std::string_view> &seen_actions, const object_store &store,
+        const ddwaf::manifest &manifest, collection_cache &cache,
+        const std::unordered_set<rule::ptr> &rules_to_exclude,
+        const std::unordered_map<rule::ptr, object_set> &objects_to_exclude,
+        ddwaf::timer &deadline) const;
```
cataphract, line 46: This parameter list is a bit unwieldy. At least indicate which are in and out parameters.

src/collection.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace ddwaf {
+
+// The collection cache is shared by both priority and regular collections,
+// this ensures that regular collections aren't processed when the respective
+// priority collection has already had a match.
```
cataphract, line 20: I'm not sure I understand this comment (or the related one when setting `cache.result = true`) in the implementation file. What is the "respective" (the language here) or the "equivalent" (in the .cpp file) regular collection of a priority collection? As far as I can see, they are not created in pairs. Are you just talking about when priority collections act as regular collections (when all the actions were already seen)?

cataphract (approved):

src/collection.hpp
Anilm3: A priority collection is created for rules with `on_match`, while a regular (i.e. non-priority) collection is created for rules without `on_match`, if there are rules with the same `type` field with and without `on_match` there will be one priority collection for `type` and one regular collection for `type`. However, if there is a match in the priority collection for `type` we should skip corresponding regular collection. The existence of a pair is not guaranteed, as it depends on the ruleset.

I can clarify this comment.

src/ruleset.hpp
Anilm3: It seems I had good intentions, but was betrayed by my memory.

PR #124, opened 2022-12-14
Support for glob component and arrays on object filter

Anilm3:
This PR introduces support for the glob key path component (`*`) within object filters. This has been implemented by allowing the trie traverser to keep track and descend through many nodes at the same time.  In addition, the use of the glob also allows for arrays within the key path.

### DIFF
--- a/src/exclusion/object_filter.cpp
+++ b/src/exclusion/object_filter.cpp
@@ -41,7 +41,7 @@ void iterate_object(const path_trie::traverser &filter, const ddwaf_object *obje
     while (!path_stack.empty()) {
         auto &[current_object, current_index, current_trie] = path_stack.top();
         if (!object::is_container(current_object)) {
-            DDWAF_DEBUG("This is a bug, the object in the stack is not a map");
+            DDWAF_DEBUG("This is a bug, the object in the stack is not a container");
             path_stack.pop();
             continue;
         }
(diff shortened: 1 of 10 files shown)


src/exclusion/object_filter.cpp
```diff
@@ -30,15 +31,16 @@ void iterate_object(const path_trie::traverser &filter, const ddwaf_object *obje
...
     }
+
     std::stack<std::tuple<const ddwaf_object *, unsigned, path_trie::traverser>> path_stack;
     path_stack.push({object, 0, filter});
 
     while (!path_stack.empty()) {
         auto &[current_object, current_index, current_trie] = path_stack.top();
-        if (current_object->type != DDWAF_OBJ_MAP) {
+        if (!object::is_container(current_object)) {
             DDWAF_DEBUG("This is a bug, the object in the stack is not a map");
```
cataphract, line 44: Don't think the message applies anymore as it now accepts arrays as well.

src/exclusion/object_filter.cpp
```diff
@@ -51,14 +53,15 @@ void iterate_object(const path_trie::traverser &filter, const ddwaf_object *obje
         for (; current_index < size; ++current_index) {
             ddwaf_object *child = &current_object->array[current_index];
 
+            path_trie::traverser child_traverser{nullptr};
             // Only consider children with keys
             if (child->parameterName == nullptr || child->parameterNameLength == 0) {
-                continue;
+                child_traverser = current_trie.descend();
```
cataphract, line 59: I'd rename it `descendWildcard`

src/exclusion/object_filter.hpp
```diff
@@ -82,25 +85,84 @@ class path_trie {
...
+        {
+            if (root != nullptr) {
+                cur_nodes_.emplace_back(root);
+            }
+        }
+
         [[nodiscard]] traverser descend(std::string_view next_key) const
         {
             if (get_state() != state::intermediate_node) {
-                // once found/not_found, as we descend we keep the state
```
cataphract, line 89: comment still applies, no?

cataphract (approved):

PR #126, opened 2022-12-19
Backtracking traverser

Anilm3:


src/exclusion/object_filter.hpp
```diff
@@ -85,85 +82,109 @@ class path_trie {
...
+            if (next_node == nullptr && seen_globs.empty()) {
+                return traverser{nullptr};
+            }
 
-                const auto *glob_node = cur_node->get_child("*");
-                if (glob_node != nullptr) {
-                    if (glob_node->is_terminal()) {
-                        return traverser{glob_node};
-                    }
+            auto new_stack = key_stack;
```
Anilm3, line 127: We could start updating the key stack once the first glob has been seen.

PR #132, opened 2023-01-24
Add boolean getter

estringana:


Anilm3:
This getter also needs to be added to `libddwaf.def` to be exported.

estringana:
> This getter also needs to be added to `libddwaf.def` to be exported.

Done

Anilm3 (approved):

PR #138, opened 2023-02-13
WAF Ruleset Builder & Release v1.8.0

Anilm3:
- Introduce ruleset builder to allow the dynamic generation of rulesets
- Support `rules_override`
- Introduce ruleset specification and update v2::parser to generate intermediate objects
- Introduce `ddwaf_update` for all-inclusive WAF ruleset updates
- Deprecate `ddwaf_toggle_rules` and `ddwaf_update_rule_data`
- Remove implicit mutexes (`std::atomic_shared_ptr`)
- Remove side-effects on running contexts
- Generalize rule tags
- Introduce naive multi-key map implementation

Remaining:
- Builder & WAF tests - since these are covered already by multiple layers of tests I might add these after the release.
- Validator tests - new features required to support this, so it'll be done in a future PR.


src/builder.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+    if (version != 2) {
+        DDWAF_ERROR("incompatible ruleset version %u.x", version);
+        throw unsupported_version();
+    }
+
+    return build_helper(ruleset, info);
+}
+
+namespace {
```
estringana, line 51: Out of interest, why is this within a namespace without name?

estringana (approved):
Overall looks good to me. So many things going on but nothing that would make me block the PR

src/builder.cpp
Anilm3: Unnamed / anonymous namespaces are used to enforce internal linkage, i.e. anything defined within it can only be used inside the file itself. It's equivalent to using global-static.

src/builder.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+std::set<rule::ptr> target_to_rules(const std::vector<parser::rule_target_spec> &targets,
+    const std::unordered_map<std::string_view, rule::ptr> &rules, const rule_tag_map &rules_by_tags)
+{
+    std::set<rule::ptr> rule_targets;
```
cataphract, line 37: `rule::ptr` being a shared_ptr (this could be made more explicit btw, say using `sptr` instead of `ptr`), this method has a lot of unnecessary refcount changes. There is actually no shared ownership required here.

src/builder.cpp
```diff
@@ -0,0 +1,312 @@
...
+    if (state == change_state::none) {
+        return {};
+    }
+
+    constexpr change_state rule_update =
+        change_state::rules | change_state::data | change_state::overrides;
+    constexpr change_state filters_update = rule_update | change_state::filters;
+    constexpr change_state manifest_update = change_state::rules | change_state::filters;
+
+    if ((state & rule_update) != change_state::none) {
```
cataphract, line 77: The fact that merely changing data (a relatively frequent occurrence) causes all this to be recalculated is quite problematic IMO.

src/builder.hpp
```diff
@@ -0,0 +1,75 @@
...
+    parser::rule_data_container dynamic_processors_;
+    parser::override_spec_container overrides_;
+    parser::filter_spec_container exclusions_;
+
+    std::unordered_map<std::string_view, rule::ptr> final_rules_;
+    rule_tag_map rules_by_tags_;
+    std::unordered_set<manifest::target_type> targets_from_rules_;
+
+    std::unordered_map<std::string_view, exclusion::rule_filter::ptr> rule_filters_;
+    std::unordered_map<std::string_view, exclusion::input_filter::ptr> input_filters_;
```
cataphract: This granular use of shared_ptrs is a bit on the wasteful side. For instance, take the rules. While you frequently manage to avoid having to change the refcount (when iterating taking the shared_ptr by reference), if the point is that rules, filters etc can be shared between rulesets, then it's those collections that should be behind the shared_ptr, not the individual filters and rules.

src/obfuscator.hpp
```diff
@@ -31,8 +31,8 @@ class obfuscator {
     static constexpr std::string_view default_key_regex_str{
         R"((p(ass)?w(or)?d|pass(_?phrase)?|secret|(api_?|private_?|public_?)key)|token|consumer_?(id|key|secret)|sign(ed|ature)|bearer|authorization)"};
 
-    std::unique_ptr<re2::RE2> key_regex{nullptr};
-    std::unique_ptr<re2::RE2> value_regex{nullptr};
+    std::shared_ptr<re2::RE2> key_regex{nullptr};
+    std::shared_ptr<re2::RE2> value_regex{nullptr};
```
cataphract, line 35: Same as mentioned before. It seems that would should be shared would be the obfuscator, rather than relying on the copy constructor to create a new obfuscator with shared ownership of the internal state.

cataphract (commented):
I looked more closely at builder.cpp due to limitations on my attention span. My main concerns are that previously relatively cheap operations like changing data or toggling rules now get much costly and that this effective ruleset calculation still leaves most of the work to the bindings. In fact, after this, I think that only things I would be able to remove from the binding/tracer would be:

a) applying the calculated final configuration merge of rules_overrides for toggling (at a big performance penalty; also what this is applied to includes the custom rules, which need to be merged to the base rules by the clients)
b) applying the calculated final configuration merge of rules_overrides to changing on_match.

The only other advantage is that changing filters requires less computation because no recalculation of rules is required.

But unless I'm missing something, that's it. So at the end, I'm not sure this is a net improvement.

include/ddwaf.h
```diff
@@ -216,37 +215,33 @@ typedef void (*ddwaf_log_cb)(
...
  *
- * Update existing rules with new rule data.
+ * @return Handle to the WAF instance.
  *
- * @param handle to the WAF instance.
- * @param data A ddwaf_object with the format [{id, type, [data]}].
- */
-DDWAF_RET_CODE ddwaf_update_rule_data(ddwaf_handle handle, ddwaf_object *data);
+ * @note If config is NULL, default values will be used, including the default
+ *       free function (ddwaf_object_free).
```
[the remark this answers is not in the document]
Anilm3: I updated this while you were reviewing it seems.

src/builder.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <rule.hpp>
+#include <ruleset.hpp>
+#include <ruleset_info.hpp>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace ddwaf {
+
+class builder {
```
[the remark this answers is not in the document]
Anilm3: I thought exactly the same name later on, but was too lazy to change it

src/builder.hpp
```diff
@@ -0,0 +1,75 @@
...
+    enum class change_state : uint32_t {
+        none = 0,
+        rules = 1,
+        overrides = 2,
+        filters = 4,
+        data = 8
+    };
+
+    friend constexpr change_state operator|(change_state lhs, change_state rhs);
+    friend constexpr change_state operator&(change_state lhs, change_state rhs);
```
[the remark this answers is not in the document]
Anilm3: Using `enum class` with values outside of the enumeration is a very interesting trick to create type-safe integers.

include/ddwaf.h
```diff
@@ -216,37 +215,33 @@ typedef void (*ddwaf_log_cb)(
...
+ **/
+ddwaf_handle ddwaf_update(ddwaf_handle handle, const ddwaf_object *rule,
```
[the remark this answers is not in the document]
Anilm3: This was updated while you were reviewing. 

> From this I can take that concurrent calls to ddwaf_update (even with different ddwaf_handles as they can share the same builder!) are forbidden and that ddwaf_update actually has side effects on the original handle.

Yes, indeed `ddwaf_update` isn't thread-safe regardless of the handle, I'll document it.

> Can the old handle still be used for non-update purposes? It seems to be case, because the ruleset is constructed anew.

Yes, it's still valid.

> Is synchronization needed to ensure that the memory state behind the new ddwaf_handle is fully visible on other threads when ddwaf_run is called? I guess the answer is no, so the bindings need at least an acquire/release when reading where the current ddwaf_handle is stored.

I think you'll only need to synchronize reads and writes around the shared `ddwaf_handle`.

src/builder.hpp
```diff
@@ -0,0 +1,75 @@
...
+    change_state load(parameter::map &root, ruleset_info &info);
+
+    std::shared_ptr<ruleset> build_helper(parameter::map &root, ruleset_info &info);
+
+    object_limits limits_;
+    ddwaf_object_free_fn free_fn_;
+    ddwaf::obfuscator event_obfuscator_;
+
+    manifest target_manifest_;
+    std::unordered_map<std::string, std::string> rule_data_ids_;
```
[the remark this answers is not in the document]
Anilm3: This was done while you were reviewing as well.

src/builder.cpp
Anilm3: I agree I need to change this to use raw pointers

src/builder.cpp
Anilm3: As discussed privately, I'll improve this with a dynamic rule processor map in the next iteration (1.8.1).

src/builder.cpp
```diff
@@ -0,0 +1,312 @@
...
+            throw ddwaf::parsing_error("no valid rules found");
+        }
+
+        // Upon reaching this stage, we know our base ruleset is valid
+        base_rules_ = std::move(new_base_rules);
+        rule_data_ids_ = std::move(rule_data_ids);
+        state = state | change_state::rules;
+    }
+
+    it = root.find("rules_data");
```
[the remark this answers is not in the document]
Anilm3: Not sure there's anything actionable on this comment...

src/builder.hpp
```diff
@@ -0,0 +1,75 @@
...
+        : limits_(limits), free_fn_(free_fn), event_obfuscator_(std::move(event_obfuscator))
+    {}
+
+    ~builder() = default;
+    builder(builder &&) = default;
+    builder(const builder &) = delete;
+    builder &operator=(builder &&) = default;
+    builder &operator=(const builder &) = delete;
+
+    std::shared_ptr<ruleset> build(parameter object, ruleset_info &info);
```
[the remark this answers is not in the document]
Anilm3: I don't understand what isn't descriptive about `parameter object`.... okay, fine.

src/builder.hpp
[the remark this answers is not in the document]
Anilm3: This was done while you were reviewing.

src/builder.cpp
Anilm3: This will be a slightly big change so I'll do so for 1.8.1

src/builder.hpp
Anilm3: I'll do this for 1.8.1

src/obfuscator.hpp
Anilm3: Done

src/builder.cpp
Anilm3: I managed to implement the dynamic rule processor map so the rules don't need to be recomputed any longer after a data update.

src/parser/parser_v2.cpp
```diff
@@ -427,33 +431,39 @@ rule_data_container parse_rule_data(
...
                 auto parsed_data = parser::parse_rule_data<rule_data_type>(type, data);
                 processor = std::make_shared<rule_processor::ip_match>(parsed_data);
-            } else if (it->second == "exact_match") {
+            } else if (operation == "exact_match") {
                 using rule_data_type = rule_processor::exact_match::rule_data_type;
                 auto parsed_data = parser::parse_rule_data<rule_data_type>(type, data);
                 processor = std::make_shared<rule_processor::exact_match>(parsed_data);
             } else {
-                DDWAF_WARN("Processor %s doesn't support dynamic rule data", it->second.c_str());
+                DDWAF_WARN("Processor %s doesn't support dynamic rule data", operation.data());
```
cataphract, line 467: In general, string_views need not be nul-terminated (depends on the source of the buffer).

cataphract (approved):

PR #141, opened 2023-02-17
Use raw pointers instead of shared pointers for rule targets

Anilm3:


cataphract (approved):
Looks good, just wonder if some/all of the `rule *` can be made `const rule *`.

PR #145, opened 2023-03-06
Remove a libinjection signature

Taiki-San:
Every trigger I could find are false positives


Anilm3 (approved):

Anilm3:
Ignore the fuzzer, I need to fix the action.

PR #146, opened 2023-03-09
R&D on go tracer bindings

eliottness:
Won't be merged. This is only used to keep track of the work done and run some CI jobs.


eliottness:
Since CI jobs do not runs on draft PR I new to put remove the draft mode

PR #148, opened 2023-03-15
Benchmark: context destroy fixture

Anilm3:


perf/main.cpp
```diff
@@ -39,6 +40,8 @@ std::map<std::string, benchmark::object_generator::settings> default_tests = {
     {"run.random.deep_containers", {.container_depth = {5, 20}, .type = generator_type::random}},
     {"run.valid", {.type = generator_type::valid}},
     {"run.mixed", {.type = generator_type::mixed}},
+    {"context_destroy", {.type = generator_type::mixed}},
```
estringana, line 43: Not a big deal but looks like the convention was `something.somethingelse` this breaks that.

estringana (approved):

perf/main.cpp
Anilm3: The convention was `<test>.<type of data>` but in this new test the type of data is irrelevant, at least for now.

PR #151, opened 2023-03-21
Improve allocation / deallocation performance within the context using a context_allocator

Anilm3:
This PR introduces the `context_allocator` which takes advantage of the `monotonic_buffer_resource` to reduce contention on the global /standard allocator. The polymorphic allocator provides a similar functionality, however to avoid the code and (slight) memory overhead, the context allocator is used with a static thread local memory resource which doesn't require any sort of propagation.
The use of this new allocator is limited to some of the objects contained within the context itself as the memory is only deleted on resource destruction.


src/iterator.cpp
```diff
@@ -42,7 +42,7 @@ template <typename T> std::vector<std::string> iterator_base<T>::get_current_pat
         return path_;
     }
 
-    std::vector<std::string> keys = path_;
+    memory::vector<memory::string> keys = path_;
     keys.reserve(path_.size() + stack_.size());
```
cataphract, line 46: You're causing a reallocation of the array inside the vector. Better to reserve and then copy from the two sources.

src/iterator.hpp
```diff
@@ -52,7 +53,7 @@ template <typename T> class iterator_base {
     // since the iterator doesn't keep track of the root object provided,
     // but only the beginning of the key path, we keep this here so that we
     // can later provide the accurate full key path.
-    std::vector<std::string> path_;
+    memory::vector<memory::string> path_;
```
cataphract, line 56: I don't think this should use the monotonic buffer. You can possibly create many iterators, and their memory will never be reclaimed.

Anilm3: Alright, that's fair enough.

src/context_allocator.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include <vector>
+
+namespace ddwaf::memory {
+extern thread_local std::pmr::memory_resource *local_memory_resource;
+
+inline std::pmr::memory_resource *get_local_memory_resource() { return local_memory_resource; }
+
+inline void set_local_memory_resource(std::pmr::memory_resource *mr) { local_memory_resource = mr; }
+
+class null_memory_resource final : public std::pmr::memory_resource {
```
cataphract, line 22: The standard provides `std::pmr::null_memory_resource()`.

src/memory_resource.hpp
```diff
@@ -0,0 +1,42 @@
...
+#  include <experimental/memory_resource>
+#  include <string>
+#  include <unordered_map>
+#  include <unordered_set>
+#  include <vector>
+
+namespace std { // NOLINT(cert-dcl58-cpp)
+namespace experimental::pmr {
+template <class Key, class T, class Hash = std::hash<Key>, class KeyEqual = std::equal_to<Key>>
+using unordered_map =
```
cataphract, line 23: this is not being used anymore

cataphract (approved):

estringana (approved):

PR #154, opened 2023-03-28
Custom rules support

Anilm3:
Custom rules support, some facts for the reviewer:
- Custom rules live alongside standard rules.
- Custom rules have precedence over ASM rules, however rules with side-effects still have precedence over rules without side-effects.
- Custom rules are sufficient to instantiate the WAF.
- Exclusion filters can affect custom rules.
- Rule overrides can't affect custom rules.


src/parser/parser_v2.cpp
```diff
@@ -501,6 +502,7 @@ override_spec_container parse_overrides(parameter::vector &override_array)
 
     return overrides;
 }
+#include <parser/specification.hpp>
```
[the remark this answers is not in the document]
Anilm3: Weird, I have no idea how that got there so I'm going to blame clang-format :-)

estringana (approved):

src/context.cpp
```diff
@@ -138,15 +138,27 @@ memory::vector<event> context::match(const memory::unordered_set<rule *> &rules_
...
+    // Evaluate user priority collections first
+    for (auto &[type, collection] : ruleset_->user_priority_collections) {
+        DDWAF_DEBUG("Evaluating user priority collection %s", type.data());
+        eval_collection(type, collection);
+    }
+
     // Evaluate priority collections first
-    for (auto &[type, collection] : ruleset_->priority_collections) {
+    for (auto &[type, collection] : ruleset_->base_priority_collections) {
         DDWAF_DEBUG("Evaluating priority collection %s", type.data());
```
cataphract, line 149: I'm not sure about these logging statements. It's unclear if `type.data()` is always nul terminated or even if `data` is not null. In particular, `get_tag` can return an empty `string_view`.

src/rule.hpp
```diff
@@ -77,6 +82,7 @@ class rule {
     }
 
     bool enabled{true};
+    source_type source{source_type::base};
```
cataphract, line 85: you already initialize it in the constructors

src/ruleset.hpp
```diff
@@ -57,12 +59,15 @@ struct ruleset {
     std::unordered_map<std::string_view, exclusion::input_filter::ptr> input_filters;
 
     // Rules are ordered by rule.id
```
cataphract, line 61: comment looks stale

src/ruleset.hpp
```diff
@@ -57,12 +59,15 @@ struct ruleset {
...
-    std::unordered_map<std::string_view, rule::ptr> rules;
+    std::vector<rule::ptr> rules;
     std::unordered_map<std::string, rule_processor::base::ptr> dynamic_processors;
 
     // Both collections are ordered by rule.type
```
cataphract, line 65: here too: "the key is rule.type"

src/ruleset_builder.cpp
```diff
@@ -36,6 +36,10 @@ namespace {
 std::set<rule *> target_to_rules(const std::vector<parser::rule_target_spec> &targets,
     const std::unordered_map<std::string_view, rule::ptr> &rules, const rule_tag_map &rules_by_tags)
 {
+    if (rules_by_tags.empty()) {
+        return {};
+    }
```
cataphract, line 41: Does this make sense? The `else` branch doesn't use `rules_by_tags`. Or is it that tags are mandatory and so there are no rules if the map is empty? It could be clearer.

src/ruleset_builder.cpp
```diff
@@ -202,15 +242,39 @@ ruleset_builder::change_state ruleset_builder::load(parameter::map &root, rulese
...
+        if (!rules.empty()) {
+            auto new_user_rules = parser::v2::parse_rules(
+                rules, info, target_manifest_, rule_data_ids, limits_, rule::source_type::user);
+
+            if (new_user_rules.empty()) {
+                DDWAF_WARN("No valid custom rules found");
+            } else {
+                // Upon reaching this stage, we know our base ruleset is valid
+                user_rules_ = std::move(new_user_rules);
+                state = state | change_state::custom_rules;
```
cataphract, line 269: If I'm reading this right, there's kind of an asymmetry in that if we already have rules, these are replaced if we have one invalid rule and one valid rule, but not if we have only one invalid rule. IMO, the old rules should be discarded in both cases.

cataphract (approved):

src/context.cpp
Anilm3: I think we've discussed this before as well...

src/ruleset_builder.cpp
Anilm3: I'm not sure why I added this, rules should always have at least type and category so I'll just remove it.

src/ruleset_builder.cpp
Anilm3: Thinking about it, I guess that's fair, otherwise it might be difficult to understand the state of the WAF.

src/context.cpp
Anilm3: To clarify, the `type` tag always needs to be present as it's enforced by the parser, so the only real concern is the nul-termination.

Anilm3: Using `%.*s` format now.

src/ruleset_builder.cpp
Anilm3: I've fixed this asymmetry for all top-level keys.

PR #157, opened 2023-03-30
clock: add vdso support for aarch64

eliottness:
This PR adds support to load the vDSO version of `clock_gettime` for aarch64 as described in [`man vdso`](https://man7.org/linux/man-pages/man7/vdso.7.html).


Anilm3 (approved):
lgtm

PR #160, opened 2023-04-05
Add all rule tags to event

Anilm3:
- Add all existing rule tags to JSON event
- Use proper access specifier for rule members
- Avoid string copies and intermediate structures by adding a pointer to the rule that generated the event within the event itself.


estringana (approved):

src/event.cpp
```diff
@@ -140,13 +154,16 @@ void event_serializer::serialize(const memory::vector<event> &events, ddwaf_resu
...
-            output.actions.array = static_cast<char **>(malloc(sizeof(char *) * actions.size()));
-            output.actions.size = actions.size();
+            output.actions.array =
+                static_cast<char **>(malloc(sizeof(char *) * all_actions.size()));
+            output.actions.size = all_actions.size();
 
             std::size_t index = 0;
-            for (const auto &action : actions) { output.actions.array[index++] = to_cstr(action); }
+            for (const auto &action : all_actions) {
+                output.actions.array[index++] = to_cstr(action);
```
cataphract, line 165: you could save some allocations here by not using to_cstr() by adding to all_actions a string_view created from std::string.c_str(), using the std::string in rule.actions

cataphract (approved):

src/event.cpp
Anilm3: As you mentioned privately, these need to outlive the ruleset.

PR #170, opened 2023-05-10
Support for per-input transformers

Anilm3:
Support per-input transformers with the schema of the following example:

```
"inputs": [
  {
    "address": "http.client_ip",
    "key_path": [],
    "transformers": []
  }
],

### DIFF
--- a/src/parser/parser_v2.cpp
+++ b/src/parser/parser_v2.cpp
@@ -100,9 +100,38 @@ std::pair<std::string, rule_processor::base::ptr> parse_processor(
     return {std::move(rule_data_id), std::move(processor)};
 }
 
+std::vector<PW_TRANSFORM_ID> parse_transformers(
+    const parameter::vector &root, condition::data_source &source)
+{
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: src/parser/parser_v2.cpp [hunk 1]
```diff
@@ -136,12 +165,18 @@ condition::ptr parse_rule_condition(const parameter::map &root, manifest &target
         target.root = target_manifest.insert(address);
         target.name = address;
         target.key_path = std::move(kp);
+        auto input_transformers = at<parameter::vector>(input, "transformers", {});
+        if (input_transformers.empty()) {
+            target.transformers = transformers;
+        } else {
+            target.transformers = parse_transformers(input_transformers, source);
```
cataphract, line 172: You don't seem to be handling `source` correctly. FWICT `parse_transformers` takes the source by reference, so it will change the default for subsequent inputs. It will also affect the value used when creating the `condition`.

cataphract (approved):

PR #171, opened 2023-05-17
Fix benchmark, test output and update ruleset to 1.7.0

Anilm3:
- Fix benchmark `--test` option
- Add boolean support to YAML emitter in benchmark
- Fix test output to show formatted events rather than JSON string
- Update reference ruleset to 1.7.0


Anilm3:
Non-critical PR (tests + benchmark), merging without approval.

PR #178, opened 2023-06-22
Refactor transformers

Anilm3:
First iteration of transformer refactoring, the objectives of this refactor:
- First and foremost, ensure no transformer can modify the original string / object.
- Avoid, but still support, two-passes whenever possible.
- Improve test coverage and reorganise the code.

Only a subset of the transformers has been ported, the remaining ones will be ported in a subsequent PR. Also note that the new transformers are not currently in use, once all transformers are available, they will be fully integrated.

To aid the review process, you can check the original implementation in `PWTransformer.cpp` and compare it with the revised implementation, they are mostly equivalent.


src/transformer/lazy_string.hpp
```diff
@@ -0,0 +1,129 @@
...
+    {
+        if (modified_) {
+            // NOLINTNEXTLINE(hicpp-no-malloc,cppcoreguidelines-no-malloc)
+            free(buffer_);
+        }
+    }
+
+    [[nodiscard]] constexpr const char &at(std::size_t idx) const { return buffer_[idx]; }
+
+    char &operator[](std::size_t idx)
```
cataphract, line 43: nitpicking but it seems you use `at` much more often, so maybe you could use `operator[]` for that one instead

src/transformer/remove_comments.cpp
```diff
@@ -0,0 +1,107 @@
...
+            ++read;
+        }
+
+        if (type == comment_type::unknown || type == comment_type::eol) {
+            break;
+        }
+
+        if (!str.modified()) {
+            // Force a copy
+            str.finalize();
```
cataphract, line 67: I don't get why this is needed.

src/transformer/utf8.cpp
```diff
@@ -233,8 +234,8 @@ size_t normalize_codepoint(uint32_t codepoint, int32_t *wbBuffer, size_t wbBuffe
                             UTF8PROC_STRIPMARK | UTF8PROC_STRIPNA | UTF8PROC_CASEFOLD),
         nullptr);
 
-    // This decomposition is unfortunately not complete. It leaves behind a few chars like ı (i)
-    //  Moreover, some conversions like ß (ss) require casefolding, which changes the case of
+    // This decomposition is unfortunately not complete. It leaves behind a few chars like Ä± (i)
+    //  Moreover, some conversions like Ã (ss) require casefolding, which changes the case of
```
cataphract, line 238: the comments were corrupted

src/transformer/utf8.cpp
```diff
@@ -256,7 +257,7 @@ size_t normalize_codepoint(uint32_t codepoint, int32_t *wbBuffer, size_t wbBuffe
             }
         } else {
             // We're forcing a case conversion back and forth if necessary to catch the few
-            // stragglers like ı (i)
+            // stragglers like Ä± (i)
```
cataphract, line 260: idem

src/transformer/utf8.hpp
```diff
@@ -17,6 +20,8 @@ uint8_t write_codepoint(uint32_t codepoint, char *utf8_buffer, uint64_t length_l
 uint32_t fetch_next_codepoint(const char *utf8_buffer, uint64_t &position, uint64_t length);
 
 size_t normalize_codepoint(uint32_t codepoint, int32_t *wb_buffer, size_t wb_buffer_length);
+
 bool normalize_string(char **utf8_buffer, uint64_t &buffer_length);
+bool normalize_string(lazy_string &str);
 
 } // namespace ddwaf::utf8
```
cataphract, line 27: should this be in src/transformer? there's no transformer class here

src/transformer/lazy_string.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include <cstdlib>
+#include <cstring>
+#include <ddwaf.h>
+#include <stdexcept>
+#include <string_view>
+
+namespace ddwaf {
+
+class lazy_string {
```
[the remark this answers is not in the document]
Anilm3: `cow_string` was the name I used on my first commit, but I wasn't sure about the potential confusion with ruminants.

src/transformer/lazy_string.hpp
```diff
@@ -0,0 +1,129 @@
...
+public:
+    explicit lazy_string(std::string_view original)
```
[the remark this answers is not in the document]
Anilm3: I will add some more tests to make sure I haven't made any mistakes, but transformers don't require nul-terminated strings.

src/transformer/lazy_string.hpp
```diff
@@ -0,0 +1,129 @@
...
+    lazy_string(const lazy_string &) = delete;
+    lazy_string &operator=(const lazy_string &) = delete;
+    lazy_string(lazy_string &&other) = delete;
+    lazy_string &operator=(lazy_string &&other) = delete;
+
+    ~lazy_string()
+    {
+        if (modified_) {
+            // NOLINTNEXTLINE(hicpp-no-malloc,cppcoreguidelines-no-malloc)
+            free(buffer_);
```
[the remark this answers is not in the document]
Anilm3: The main reason for using `malloc` / `free` is because the `ddwaf_object` interface uses `malloc` / `free` and the ownership of all strings is transferred to a `ddwaf_object` and consequently deleted through `ddwaf_object_free`. This will all change at some point but not in this PR.

src/transformer/lazy_string.hpp
```diff
@@ -0,0 +1,129 @@
...
+        for (std::size_t i = start; i < length_; ++i) {
+            if (buffer_[i] == c) {
+                return {true, i};
+            }
+        }
+        return {false, 0};
+    }
+
+    // Reset the internal buffer, ownership is transferred
+    void reset(char *str, std::size_t length)
```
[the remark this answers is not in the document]
Anilm3: The `lazy_string` / `cow_string` is a very ad-hoc structure tailored around the transformers, this particular function was introduced for transformers which allocate their own strings, so a `unique_ptr` would be a bit odd to use here.

I agree that `reset` is confusing due to how STL types work, so I'll rename it to `replace_buffer`.

src/transformer/lazy_string.hpp
Anilm3: For consistency with `std::string`, I prefer to use `[]` to get a reference.

src/transformer/utf8.cpp
Anilm3: Vim doesn't like unicode...

src/transformer/utf8.hpp
Anilm3: These are just `utf8` utility functions only used by the transformers so I prefer to keep it here, I can put it in a `transformer/common` directory if it makes things cleaner.

src/transformer/remove_comments.cpp
Anilm3: This is an optimisation for the first iteration, when a string has no comments:
- The previous loop finds comments, but it's also used to write text before/between/after comments.
- If the string hasn't been modified, it isn't written to within that loop.
- Once we reach this point, we know there is a comment, so we force a copy here so that the next iteration enables line 53.

This essentially allows for the check to be done at the same time as the transformation, making this a single-pass transformer.

src/transformer/lazy_string.hpp
```diff
@@ -0,0 +1,129 @@
...
+            modified_ = false;
+            buffer_ = nullptr;
+            length_ = 0;
+            return res;
+        }
+        return {nullptr, 0};
+    }
+
+    // Update length and nul-terminate, allocate if not allocated
+    void finalize() { return finalize(length_); }
```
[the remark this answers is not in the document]
Anilm3: I've managed to remove `finalize`, renamed the other overload to truncate and I'm also not using `modified` anymore but kept it for testing purposes.

src/transformer/remove_comments.cpp
Anilm3: I have simplified this to make it a bit more clear.

src/transformer/lazy_string.hpp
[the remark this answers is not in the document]
Anilm3: Done

src/transformer/lazy_string.hpp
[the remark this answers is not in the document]
Anilm3: I've renamed the function to `replace_buffer`.

src/transformer/lazy_string.hpp
```diff
@@ -0,0 +1,129 @@
...
+        modified_ = true;
+        buffer_ = str;
+        length_ = length;
+    }
+
+    // Moves the contents and invalidates the string if the buffer has been
+    // modified, otherwise it does nothing
+    std::pair<char *, std::size_t> move()
+    {
+        if (modified_) {
```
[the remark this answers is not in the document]
Anilm3: Done

src/transformer/lazy_string.hpp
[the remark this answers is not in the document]
Anilm3: Done

tests/transformer/cow_string_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+    EXPECT_TRUE(str.modified());
+    EXPECT_NE(str.data(), nullptr);
+
+    auto [buffer, length] = str.move();
+    EXPECT_STREQ(buffer, "valee");
+    EXPECT_EQ(length, 5);
+    free(buffer);
+
+    EXPECT_EQ(str.length(), 0);
+    EXPECT_FALSE(str.modified());
```
estringana, line 78: Why is this false?

tests/transformer/cow_string_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+    for (size_t i = 0; i < original.length(); ++i) { EXPECT_EQ(original[i], str.at(i)); }
+
+    EXPECT_FALSE(str.modified());
+    EXPECT_NE(str.data(), nullptr);
+}
+
+TEST(TestCoWString, NonConstRead)
+{
+    constexpr std::string_view original = "value";
+    cow_string str(original);
```
estringana, line 26: I will say that nothing is testing that a `cow_string` will assert `EXPECT_FALSE(str.modified())` when created

estringana: At least explicitly

src/transformer/common/utf8.cpp
```diff
@@ -11,8 +11,8 @@
 #include <unordered_map>
```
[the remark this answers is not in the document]
Anilm3: It's not a transformer but some helper functions that are used by multiple transformers, it's indirectly tested through each of the transformers using them.

tests/transformer/cow_string_test.cpp
Anilm3: Because calling `move()` invalidates the original string.

tests/transformer/remove_comments_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+    EXPECT_TRANSFORM(remove_comments, "--", "");
+    EXPECT_TRANSFORM(remove_comments, "--foo", "");
+    EXPECT_TRANSFORM(remove_comments, "bar--foo", "bar");
+    EXPECT_TRANSFORM(remove_comments, "bar--", "bar");
+}
+
+TEST(TestRemoveComments, ValidTransformCPPComment)
+{
+    EXPECT_TRANSFORM(remove_comments, "//", "");
```
estringana, line 37: I don't have the whole background of transformers but would it be possible to get something like `//blaaaa\nNoCommentOnThisLine` ?

src/transformer/remove_comments.cpp
```diff
@@ -0,0 +1,106 @@
...
+    enum class comment_type { unknown, html, c, eol } type = comment_type::unknown;
+
+    bool comment_found = false;
+
+    std::size_t read = 0;
+    std::size_t write = 0;
+    while (read < str.length()) {
+        type = comment_type::unknown;
+        while (read < str.length()) {
+            if (str.at(read) == '<' && read + 3 < str.length() && str.at(read + 1) == '!' &&
```
estringana, line 22: Since all strings on this and on every other transformers are `cow_string` would it be possible to have a method to avoid having to check things like `read + x < lengh && read + x == Y`.

tests/transformer/cow_string_test.cpp
estringana: Why not on this one?

estringana (approved):

tests/transformer/remove_comments_test.cpp
Anilm3: In theory yes, but the remove comments transformer doesn't cover that case and I'm not sure if we want to cc: @Taiki-San

src/transformer/common/utf8.cpp
```diff
@@ -361,4 +362,81 @@ bool normalize_string(char **_utf8Buffer, uint64_t &bufferLength)
     return true;
 }
 
+bool normalize_string(cow_string &str)
```
cataphract, line 365: the other function could be written in function of this, since they're basically the same

cataphract (approved):

src/transformer/common/utf8.cpp
Anilm3: The other function will be removed in the next iteration, as it'll deprecate `PWTransformer.*`, so I decided to be pragmatic.

PR #184, opened 2023-07-10
Rule filter bypass / monitor mode support

Anilm3:
This PR adds support for rule filter monitor and bypass (default) modes:
- In monitor mode, rules are still evaluated but actions (`on_match`) are not provided as part of the result.
- Bypass mode  is the current (and default) behaviour in which a rule is simply skipped from evaluation.

### DIFF
--- a/src/exclusion/rule_filter.hpp
+++ b/src/exclusion/rule_filter.hpp
@@ -15,6 +15,9 @@
 #include <rule.hpp>
 
 namespace ddwaf::exclusion {
+
+enum class filter_mode { bypass, monitor };
+
 class rule_filter {
 public:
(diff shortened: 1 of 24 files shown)


estringana (approved):

src/exclusion/rule_filter.hpp
```diff
@@ -43,6 +47,7 @@ class rule_filter {
     std::string id_;
     std::vector<condition::ptr> conditions_;
     std::unordered_set<rule *> rule_targets_;
+    filter_mode mode_{filter_mode::bypass};
```
cataphract, line 50: already initialized in the constructor

cataphract (approved):

PR #188, opened 2023-07-27
Send monitor as event action instead of block

estringana:


src/event.cpp
```diff
@@ -105,7 +105,12 @@ void event_serializer::serialize(const memory::vector<event> &events, ddwaf_resu
                     if (!event.skip_actions) {
                         all_actions.emplace(action);
                     }
-                    ddwaf_object_array_add(&actions_array, to_object(tmp, action));
+
+                    if (event.skip_actions && action == "block") {
+                        ddwaf_object_array_add(&actions_array, to_object(tmp, "monitor"));
```
Anilm3, line 110: This is inside a loop so if we have 10 actions, we'll have 10 `monitor`.

Anilm3 (changes requested):

Anilm3 (approved):

PR #194, opened 2023-08-02
Equals operator and arbitrary operator type support

Anilm3:
- Add support  for arbitrary types on operations
- Refactor operations and use CRTP to simplify implementation
- Introduce `equals` operator, supporting `string`, `int64`, `uint64` and `bool`
- Remove `std::to_string` and replace with custom `ddwaf::to_string<StringType>` since it's not thread-safe, locale dependent and doesn't support custom allocators:

> std::to_string relies on the current locale for formatting purposes, and therefore concurrent calls to std::to_string from multiple threads may result in partial serialization of calls. C++17 provides [std::to_chars](https://en.cppreference.com/w/cpp/utility/to_chars) as a higher-performance locale-independent alternative.

- Remove some compiler-generated branches on coverage (hence the jump)
- Use `unique_ptr` operations in `expression` and `shared_ptr` in dynamic operations


src/operation/base.hpp
```diff
@@ -19,35 +19,79 @@ namespace ddwaf::operation {
...
     {
-        return {memory::string(resolved), memory::string(matched), name(), to_string(), {}, {}};
+        return static_cast<const T *>(this)->match_impl(data);
+    }
+
+    [[nodiscard]] std::pair<bool, memory::string> match(const ddwaf_object &obj) const override
+    {
+        const auto *ptr = static_cast<const T *>(this);
+        if constexpr (T::supported_type_impl() == DDWAF_OBJ_STRING) {
+            if (obj.stringValue == nullptr) {
```
cataphract, line 77: you're looking at the union value without checking the union type tag. idk if this is checked by the caller, but even if it is, this being a public a method it shouldn't be unsafe

src/operation/equals.hpp
```diff
@@ -0,0 +1,72 @@
...
+    equals(const equals &) = default;
+    equals(equals &&) noexcept = default;
+    equals &operator=(const equals &) = default;
+    equals &operator=(equals &&) noexcept = default;
+
+protected:
+    static constexpr std::string_view to_string_impl() { return ""; }
+    static constexpr std::string_view name_impl() { return "equals"; }
+
+    template <typename U = T, std::enable_if_t<std::is_same<U, int64_t>::value, bool> = true>
```
cataphract, line 32: could be a bit simplified with

```c++
    template <typename U = T, typename = std::enable_if_t<std::is_same_v<U, int64_t>>>
```

or since c++20 is enabled...

```c++
    static constexpr DDWAF_OBJ_TYPE supported_type_impl() requires std::is_same_v<T, int64_t>
    {
        return DDWAF_OBJ_SIGNED;
    }
```

but still my preference would be a single definition and `if constexpr`.

Anilm3: Ah... the burden of choice!

src/operation/base.hpp
Anilm3: Right, I'll add a check.

src/expression.cpp
```diff
@@ -11,41 +11,79 @@
...
+    }
+
+    return {};
+}
+
+} // namespace
+
 std::optional<event::match> expression::evaluator::eval_object(const ddwaf_object *object,
-    const operation::base::ptr &processor, const std::vector<transformer_id> &transformers) const
+    const operation::base *processor, const std::vector<transformer_id> &transformers) const
```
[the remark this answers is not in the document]
Anilm3: Makes sense although `get_processor` still needs to return a pointer, but that's okay.

src/operation/equals.hpp
Anilm3: I'll use `if constexpr` since it's more compact and I've used it in other places as well in this PR.

cataphract (approved):

src/exclusion/object_filter.hpp
```diff
@@ -165,9 +165,9 @@ class path_trie {
         std::vector<const trie_node *> cur_nodes_;
     };
 
-    template <typename StringType,
-        typename = std::enable_if<std::is_constructible<std::string, StringType>::value>>
+    template <typename StringType>
     void insert(const std::vector<StringType> &path)
+        requires std::is_constructible_v<std::string, StringType>
```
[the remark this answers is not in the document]
Anilm3: Great catch

PR #197, opened 2023-08-03
Float and null type support

Anilm3:
Add support for `float` and `null` types, while also allowing `invalid` within containers.

Breaking changes:

- `ddwaf_object` now has one more field in the union for doubles called `f64`
- `ddwaf_object_*_force` are now the standard, so the `_force` suffix has been removed.
- `ddwaf_object_signed` and `ddwaf_object_unsigned` have now been renamed to `ddwaf_object_string_from_*`
- Not breaking but two new types have been introduced, `DDWAF_OBJ_FLOAT` and `DDWAF_OBJ_NULL`

### DIFF
--- a/src/matcher/equals.hpp
+++ b/src/matcher/equals.hpp
@@ -9,6 +9,7 @@
 #include <clock.hpp>
 #include <matcher/base.hpp>
 #include <string_view>
+#include <type_traits>
 #include <unordered_map>
 #include <utils.hpp>
 
@@ -18,7 +19,10 @@ template <typename T> class equals : public base_impl<equals<T>> {
 public:
     using rule_data_type = std::vector<std::pair<std::string_view, uint64_t>>;
 
-    explicit equals(T expected) : expected_(std::move(expected)) {}
+    explicit equals(T expected)
+        requires(!std::is_floating_point_v<T>)
+        : expected_(std::move(expected))
+    {}
     ~equals() override = default;
     equals(const equals &) = default;
     equals(equals &&) noexcept = default;
@@ -62,6 +66,34 @@ protected:
     T expected_;
 
     friend class base_impl<equals<T>>;
+};
+
+template <> class equals<double> : public base_impl<equals<double>> {
+public:
+    using rule_data_type = std::vector<std::pair<std::string_view, uint64_t>>;
+
+    // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
+    equals(double expected, double precision) : expected_(expected), precision_(precision) {}
+    ~equals() override = default;
+    equals(const equals &) = default;
+    equals(equals &&) noexcept = default;
+    equals &operator=(const equals &) = default;
+    equals &operator=(equals &&) noexcept = default;
+
+protected:
+    static constexpr std::string_view to_string_impl() { return ""; }
+    static constexpr std::string_view name_impl() { return "equals"; }
+    static constexpr DDWAF_OBJ_TYPE supported_type_impl() { return DDWAF_OBJ_FLOAT; }
+
+    [[nodiscard]] std::pair<bool, memory::string> match_impl(double obtained) const
+    {
+        return {std::abs(expected_ - obtained) < precision_, {}};
+    }
+
+    double expected_;
+    double precision_;
+
+    friend class base_impl<equals<double>>;
 };
 
 } // namespace ddwaf::matcher
--- a/src/matcher/equals.hpp
+++ b/src/matcher/equals.hpp
@@ -73,7 +73,7 @@ public:
     using rule_data_type = std::vector<std::pair<std::string_view, uint64_t>>;
 
     // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
-    equals(double expected, double precision) : expected_(expected), precision_(precision) {}
+    equals(double expected, double delta) : expected_(expected), delta_(delta) {}
     ~equals() override = default;
     equals(const equals &) = default;
     equals(equals &&) noexcept = default;
@@ -87,11 +87,11 @@ protected:
 
     [[nodiscard]] std::pair<bool, memory::string> match_impl(double obtained) const
     {
-        return {std::abs(expected_ - obtained) < precision_, {}};
+        return {std::abs(expected_ - obtained) < delta_, {}};
     }
 
     double expected_;
-    double precision_;
+    double delta_;
 
     friend class base_impl<equals<double>>;
 };
--- a/src/matcher/equals.hpp
+++ b/src/matcher/equals.hpp
@@ -17,8 +17,6 @@ namespace ddwaf::matcher {
 
 template <typename T> class equals : public base_impl<equals<T>> {
 public:
-    using rule_data_type = std::vector<std::pair<std::string_view, uint64_t>>;
-
     explicit equals(T expected)
         requires(!std::is_floating_point_v<T>)
         : expected_(std::move(expected))
@@ -70,8 +68,6 @@ protected:
 
 template <> class equals<double> : public base_impl<equals<double>> {
 public:
-    using rule_data_type = std::vector<std::pair<std::string_view, uint64_t>>;
-
     // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
     equals(double expected, double delta) : expected_(expected), delta_(delta) {}
     ~equals() override = default;
--- a/src/object.cpp
+++ b/src/object.cpp
@@ -238,8 +238,8 @@ bool ddwaf_object_array_add(ddwaf_object *array, ddwaf_object *object)
         DDWAF_DEBUG("Invalid call, this API can only be called with an array as first parameter");
         return false;
     }
-    if (object == nullptr || object->type == DDWAF_OBJ_INVALID) {
-        DDWAF_DEBUG("Tried to add an invalid entry to an array");
+    if (object == nullptr) {
+        DDWAF_DEBUG("Tried to add a null object to an array");
         return false;
     }
     return ddwaf_object_insert(array, *object);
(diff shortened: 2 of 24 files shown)


src/utils.hpp
```diff
@@ -127,20 +133,37 @@ template <class Fn> class scope_exit {
...
+StringType to_string(T value)
+    requires std::is_floating_point_v<T>
+{
+    using char_type = typename StringType::value_type;
+    using traits_type = typename StringType::traits_type;
+    using allocator_type = typename StringType::allocator_type;
+    std::basic_ostringstream<char_type, traits_type, allocator_type> ss;
+    ss << value;
+    return std::move(ss).str();
+}
```
cataphract, line 160: Use `std::to_chars`. The performance is an order of magnitude better:

```c++
template <typename StringType, typename T>
StringType to_string(T value)
    requires std::is_floating_point_v<T> &&
             std::is_same_v<
                 StringType,
                 std::basic_string<typename StringType::value_type,
                                   typename StringType::traits_type,
                                   typename StringType::allocator_type>>
{
    using char_type = typename StringType::value_type;
    static_assert(std::is_same_v<char_type, char>);

    static constexpr std::size_t buffer_size =
        sizeof(value) == 16 ? 35 : (sizeof(value) == 8 ? 17 : 9);

    char buffer[buffer_size];
    auto result = std::to_chars(buffer, buffer + sizeof(buffer), value);
    return {buffer, result.ptr};
}
```

src/matcher/equals.hpp
```diff
@@ -64,4 +68,32 @@ template <typename T> class equals : public base_impl<equals<T>> {
     friend class base_impl<equals<T>>;
 };
 
+template <> class equals<double> : public base_impl<equals<double>> {
+public:
+    using rule_data_type = std::vector<std::pair<std::string_view, uint64_t>>;
+
+    // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
+    equals(double expected, double precision) : expected_(expected), precision_(precision) {}
```
cataphract, line 76: I'd call it _delta_ or _dx_, because the usual meaning "precision"  when dealing with quantities is the effective number of digits in the numbers. What you have here is the "absolute uncertainty".

src/utils.hpp
Anilm3: Thanks, I was using `to_chars` before but libc++ only supports `double` from version 14, so I'm working on a version using concepts to switch between each implementation.

src/matcher/equals.hpp
Anilm3: You're right, precision in this context is not the right name for it, I'll rename it to delta.

src/object.cpp
```diff
@@ -95,7 +106,7 @@ ddwaf_object *ddwaf_object_stringl_nc(ddwaf_object *object, const char *string,
     return object;
 }
 
-ddwaf_object *ddwaf_object_signed(ddwaf_object *object, int64_t value)
+ddwaf_object *ddwaf_object_string_from_signed(ddwaf_object *object, int64_t value)
```
estringana, line 109: NP: I like this new naming `from` but it feels a bit inconsistent

src/object.cpp
```diff
@@ -393,6 +417,15 @@ int64_t ddwaf_object_get_signed(const ddwaf_object *object)
     return object->intValue;
 }
 
+double ddwaf_object_get_float(const ddwaf_object *object)
+{
+    if (object == nullptr || object->type != DDWAF_OBJ_FLOAT) {
+        return 0;
```
estringana, line 423: How do user of this function know if the 0 comes from the object or from this conditional?

src/utils.hpp
```diff
@@ -125,22 +131,88 @@ template <class Fn> class scope_exit {
...
+    requires std::is_integral_v<T> &&
+             (!std::is_same_v<T, bool>) && std::is_same_v<typename StringType::value_type, char> &&
+             std::is_same_v<StringType,
+                 std::basic_string<typename StringType::value_type,
+                     typename StringType::traits_type, typename StringType::allocator_type>>
 {
     // Maximum number of characters required to represent a 64 bit integer as a string
     // 20 bytes for UINT64_MAX or INT64_MIN + null byte
-    static constexpr size_t uint64_max_chars = 21;
+    static constexpr size_t max_chars = 21;
```
estringana, line 150: I think this name now is too dependant on the comment

estringana: Acually its scope is quite small

estringana (approved):
Just a few nitpicks

src/utils.hpp
```diff
@@ -90,6 +91,11 @@ inline bool is_scalar(const ddwaf_object *obj)
     return obj != nullptr && (obj->type & PWI_DATA_TYPES) != 0;
 }
 
+inline bool is_invalid_or_null(const ddwaf_object *obj)
+{
+    return obj != nullptr && obj->type == DDWAF_OBJ_INVALID && obj->type == DDWAF_OBJ_NULL;
```
cataphract, line 96: you mean `obj != nullptr && (obj->type == DDWAF_OBJ_INVALID || obj->type == DDWAF_OBJ_NULL)` I guess

src/utils.hpp
```diff
@@ -125,22 +131,88 @@ template <class Fn> class scope_exit {
...
+concept has_from_chars = requires(T v) { std::from_chars(nullptr, nullptr, std::declval<T>()); };
+
 template <typename StringType, typename T>
 StringType to_string(T value)
-    requires std::is_integral_v<T>
+    requires std::is_integral_v<T> &&
+             (!std::is_same_v<T, bool>) && std::is_same_v<typename StringType::value_type, char> &&
+             std::is_same_v<StringType,
+                 std::basic_string<typename StringType::value_type,
+                     typename StringType::traits_type, typename StringType::allocator_type>>
```
cataphract, line 146:
```
    requires std::is_integral_v<T> &&
             !std::is_same_v<T, bool> &&
             std::is_same_v<StringType,
                 std::basic_string<char,
                     typename StringType::traits_type, typename StringType::allocator_type>>
```

src/utils.hpp
cataphract, line 150: you don't need the nul byte though

src/utils.hpp
```diff
@@ -125,22 +131,88 @@ template <class Fn> class scope_exit {
...
+    [[unlikely]] if (ec != std::errc()) {
+        return {};
+    }
+    return {str.data(), ptr};
+}
+
+template <typename StringType, typename T>
+StringType to_string(T value)
+    requires std::is_floating_point_v<T> && std::is_same_v<typename StringType::value_type, char> &&
+             std::is_same_v<StringType,
```
cataphract, line 163: same here, you can remove `std::is_same_v<typename StringType::value_type, char>` and replace `typename StringType::value_type` with `char`

src/utils.hpp
```diff
@@ -125,22 +131,88 @@ template <class Fn> class scope_exit {
...
+                 std::basic_string<typename StringType::value_type,
+                     typename StringType::traits_type, typename StringType::allocator_type>>
+{
+    if constexpr (has_to_chars<double>) {
```
cataphract, line 167: you accept T to be floating point, but then only handle the double case here. You could check `has_to_chars(T)` instead.

`max_chars` could be calculated thus:

```cpp
template<typename T>
requires std::is_same_v<T, float> || std::is_same_v<T, double>
// XXX: add long double, though it's tricker, we don't know if it's quad-precision or x87 80-bit long doubles
constexpr std::size_t max_exp_digits = sizeof(T) == 4 ? 2 : 4;

static constexpr std::size_t max_chars =
    std::numeric_limits<T>::digits10 + 1
    + 1 /* sign */ + 1 /* dot */ + 1 /* e */ + 1 /* exp sign */
    +  max_exp_digits<T>;
```

src/utils.hpp
```diff
@@ -125,22 +131,88 @@ template <class Fn> class scope_exit {
...
+        [[unlikely]] if (ec != std::errc()) {
+            // This is likely unreachable if the max_chars calculation is accurate
+            return {};
+        }
         return {str.data(), ptr};
+    } else {
+        using char_type = typename StringType::value_type;
+        using traits_type = typename StringType::traits_type;
+        using allocator_type = typename StringType::allocator_type;
+        std::basic_ostringstream<char_type, traits_type, allocator_type> ss;
```
cataphract, line 182: would be closer to the `to_chars` version if you used [`setprecision(std::numeric_limits<T>::digits10)`](https://en.cppreference.com/w/cpp/io/manip/setprecision)

The difference being that `to_chars` may be smaller if the representation is unambiguous with fewer digits.

src/utils.hpp
```diff
@@ -125,22 +131,88 @@ template <class Fn> class scope_exit {
...
+        return std::move(ss).str();
     }
-    return {};
 }
 
-template <typename StringType> StringType to_string(bool value) { return value ? "true" : "false"; }
+template <typename StringType, typename T>
+StringType to_string(T value)
+    requires std::is_same_v<T, bool> && std::is_same_v<typename StringType::value_type, char> &&
+             std::is_same_v<StringType,
```
cataphract, line 191: same

cataphract (changes requested):

src/utils.hpp
Anilm3: Ah yes, I changed it without thinking...

src/object.cpp
Anilm3: How so?

src/object.cpp
Anilm3: They can't know, although they can always check if their object is null or of an incorrect type, but it's mostly a precaution for incorrect use of the API.

src/utils.hpp
Anilm3: I don't although I have an unhealthy attachment to it.

PR #200, opened 2023-08-11
Provide weak impl of missing functions in 32-bit glibc

cataphract:


Anilm3 (approved):
Great work, this should do it!

PR #210, opened 2023-08-29
Convert float to (un)signed

Anilm3:


estringana (approved):

src/parameter.cpp
```diff
@@ -132,6 +132,11 @@ parameter::operator uint64_t() const
         return intValue;
     }
 
+    static constexpr long double uint64_max = std::numeric_limits<uint64_t>::max();
+    if (type == DDWAF_OBJ_FLOAT && (f64 >= 0.0) && (f64 <= uint64_max)) {
+        return static_cast<uint64_t>(f64);
+    }
```
cataphract, line 138: I'd add `f64 == round(f64)` or `f64 == static_cast<uint64_t>(f64)` to avoid truncation of the decimal part.

src/parameter.cpp
```diff
@@ -152,6 +157,11 @@ parameter::operator int64_t() const
         return static_cast<int64_t>(uintValue);
     }
 
+    static constexpr long double int64_max = std::numeric_limits<int64_t>::max();
+    if (type == DDWAF_OBJ_FLOAT && f64 <= int64_max) {
```
cataphract, line 161: needs to check `f64 >= int64_min` to avoid undefined behavior

Anilm3: Yes, you're right I missed that.

src/parameter.cpp
Anilm3: I can add this to both but I was trying to be extra lenient...

cataphract (approved):

PR #212, opened 2023-09-05
Remove ptr typedefs

Anilm3:
For convenience and reducing characters, I introduced a  `ptr` typedef (using declaration) on certain classes which were typically used as a `shared_ptr`. Over time the use of this `ptr` typedef was extended to other classes, some as `shared_ptr` and some as `unique_ptr`, which has made it very difficult to understand the underlying pointer type.  This PR removes them to make the underlying pointer type more explicit.


estringana (approved):

estringana (approved):

cataphract (commented):
if you wanted to save typing, you could use typedefs like `uptr` and `sptr`, but this is fine too

PR #213, opened 2023-09-06
Indexer abstraction to encapsulate rule and scanner search and storage

Anilm3:
Currently scanners and rules have a companion `*_by_tags_" container which is used to search a set of rules or scanners based on their tags and values. Similarly, rules and scanners are organised in an `unordered_map` based on their ID; this container also has ownership of the objects themselves.

To avoid a case in which a container has dangling pointers due to the owning container being destroyed and to simplify the code, an indexer structure has been introduced which:

- Stores shared pointers to owned objects in an `std::vector`, which can be directly referenced and iterated.
- Stores raw pointers to objects in a map, indexing them by ID.
- Stores raw pointers to objects in an mkmap, indexing them by tags.
- Provides sufficient functionality to search based on ID and tags.

### DIFF
--- a/src/ruleset_builder.cpp
+++ b/src/ruleset_builder.cpp
@@ -33,27 +33,26 @@ constexpr ruleset_builder::change_state operator&(
 
 namespace {
 
-std::set<rule *> references_to_rules(const std::vector<parser::reference_spec> &references,
-    const std::unordered_map<std::string_view, std::shared_ptr<rule>> &rules,
-    const rule_tag_map &rules_by_tags)
+std::set<rule *> references_to_rules(
+    const std::vector<parser::reference_spec> &references, const indexer<rule> &rules)
 {
     std::set<rule *> rule_refs;
     if (!references.empty()) {
         for (const auto &ref : references) {
             if (ref.type == parser::reference_type::id) {
-                auto rule_it = rules.find(ref.ref_id);
-                if (rule_it == rules.end()) {
+                auto *rule = rules.find_by_id(ref.ref_id);
+                if (rule == nullptr) {
                     continue;
                 }
-                rule_refs.emplace(rule_it->second.get());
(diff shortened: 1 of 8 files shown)


estringana (approved):

src/ruleset_builder.cpp
```diff
@@ -98,21 +97,19 @@ std::shared_ptr<ruleset> ruleset_builder::build(parameter::map &root, base_rules
...
     if ((state & base_rule_update) != change_state::none) {
         final_base_rules_.clear();
-        base_rules_by_tags_.clear();
 
         // Initially, new rules are generated from their spec
         for (const auto &[id, spec] : base_rules_) {
             auto rule_ptr = std::make_shared<ddwaf::rule>(
                 id, spec.name, spec.tags, spec.expr, spec.actions, spec.enabled, spec.source);
 
             // The string_view should be owned by the rule_ptr
```
cataphract, line 106: comment looks stale

src/ruleset_builder.cpp
```diff
@@ -141,16 +137,14 @@ std::shared_ptr<ruleset> ruleset_builder::build(parameter::map &root, base_rules
...
     if ((state & change_state::custom_rules) != change_state::none) {
         final_user_rules_.clear();
-        user_rules_by_tags_.clear();
 
         // Initially, new rules are generated from their spec
         for (const auto &[id, spec] : user_rules_) {
             auto rule_ptr = std::make_shared<ddwaf::rule>(
                 id, spec.name, spec.tags, spec.expr, spec.actions, spec.enabled, spec.source);
 
             // The string_view should be owned by the rule_ptr
```
cataphract, line 146: idem

cataphract (dismissed):

cataphract (approved):

estringana (approved):

PR #218, opened 2023-09-26
Simplify rule evaluation

Anilm3:
This change aims to reduce the number of rules that are evaluated on every call to `ddwaf_run`. Currently the following changes have been done:
- Replace `collections` with a `multiset` ordered depending on rule priority (actions, user vs base, type).
- Remove disabled rules from ruleset.

Remaining:
- Filter rules depending on available inputs.
- Replace collections tests.
- Reinstate ruleset test.
- Add other tests.


Anilm3:
Superseded by https://github.com/DataDog/libddwaf/pull/353

PR #221, opened 2023-10-11
Address diagnostics

Anilm3:
This PR introduces information regarding addresses to the diagnostics for each of the different features. This includes both required and optional addresses, for example:
```
  "processors": {
    "loaded": [
      "processor-001"
    ],
    "failed": [],
    "errors": {},
    "addresses": {
      "required": [ ... ],
      "optional": [ ... ]
    }
  },
  ```
  
This allows the WAF user to distinguish which addresses they actually need to provide (required) from those which are opportunistically used (optionals) . It also allows distinguishing addresses which are used in potentially disabled features, for example, the processors section would provide insights into the addresses used for schema extraction:
- Required refers to the condition addresses.
- Optional refers to the mappings used to generate schemas, these are not "required" to run the processor but if available they will be used.

Unfortunately, it's not possible to distinguish which addresses correspond to the schema extraction processor, although if more processors are included in the future, this might be required.

Finally, this PR also renames `ddwaf_required_addresses` to `ddwaf_known_addresses` to better reflect the actual contents. This function now also returns optional addresses (processor mappings, object filter addresses) and required addresses (mainly condition addresses).

Related Jira [APPSEC-11125]

[APPSEC-11125]: https://datadoghq.atlassian.net/browse/APPSEC-11125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -231,19 +231,24 @@ ddwaf_handle ddwaf_update(ddwaf_handle handle, const ddwaf_object *ruleset,
 void ddwaf_destroy(ddwaf_handle handle);
 
 /**
- * ddwaf_required_addresses
+ * ddwaf_known_addresses
  *
- * Get an array of required (root) addresses. The memory is owned by the WAF and
- * should not be freed.
+ * Get an array of known (root) addresses used by rules, exclusion filters and
+ * processors. This array contains both required and optional addresses. A more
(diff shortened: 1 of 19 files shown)


Taiki-San (commented):
lgtm!

PR #224, opened 2023-10-18
Ephemeral Exclusions

Anilm3:
Support ephemeral addresses on rule and input filters. This requires:
- Discarding excluded rules if they have been excluded as a result of an ephemeral match.
- Discarding ephemeral inputs excluded.
- Discarding inputs excluded as a result of an ephemeral match.


src/context.cpp
```diff
@@ -194,25 +195,43 @@ const memory::unordered_map<rule *, context::object_set> &context::filter_inputs
...
-                auto &common_exclusion = objects_to_exclude_[rule];
-                common_exclusion.insert(exclusion->objects.begin(), exclusion->objects.end());
+                if (!exclusion->objects.ephemeral.empty()) {
+                    auto &rule_policy = exclusion_policy_.ephemeral[rule];
+                    // Bypass has precedence over monitor
+                    if (rule_policy.mode != filter_mode::bypass) {
+                        rule_policy.objects.insert(exclusion->objects.ephemeral.begin(),
+                            exclusion->objects.ephemeral.end());
+                    }
+                }
```
cataphract, line 225: again, it feels like moving the logic to something like `exclusion_policy_.add_input_exclusion(rule, object_set)` would help

src/context.hpp
```diff
@@ -67,7 +63,7 @@ class context {
         }
         return false;
     }
-    bool should_eval_filters() const
+    bool check_new_filter_targets() const
     {
         for (const auto &[target, str] : ruleset_->filter_addresses) {
```
cataphract, line 68: tidy wants `any_of`, as in:

```c++
return std::ranges::any_of(ruleset_->filter_addresses, [this](auto &&target) {
        return store_.is_new_target(std::get<const target_index>(target));
});
```

whether it's more readable is questionable though...

cataphract (approved):

src/context.cpp
```diff
@@ -194,25 +195,43 @@ const memory::unordered_map<rule *, context::object_set> &context::filter_inputs
         input_filter::cache_type &cache = it->second;
         auto exclusion = filter->match(store_, cache, deadline);
         if (exclusion.has_value()) {
+            DDWAF_DEBUG("Has value");
```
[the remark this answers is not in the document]
Anilm3: It's a stray log statement from debugging some unit tests... I'll just remove it.

src/context.cpp
Anilm3: I agree, the main reason I didn't do it is because I want to have an abstraction on top of exclusion filters which should also make it simpler to test things. However, I don't see any harm in having this function in the `exclusion::context_policy`, so I'll add it.

src/context.hpp
Anilm3: I think this didn't compile somewhere (maybe locally in my old laptop, maybe in CI, I can't recall), so I'm ignoring it for now.

Anilm3: Okay I tried, let's see if it compiles...

Anilm3: Indeed it failed on macos and the old docker builds.

cataphract (approved):

PR #226, opened 2023-10-23
Use `fmt::format` for logging and vendorize some dependencies within `src/`

Anilm3:
Changes in this PR:
- Vendorize some dependencies within `src/vendor` and build as part of `libddwaf_objects`. This is limited to small dependencies which aren't updated often.
- Use `fmt::format` to simplify log format string.


CMakeLists.txt
```diff
@@ -118,6 +118,15 @@ set(LIBDDWAF_SOURCE
...
     ${libddwaf_SOURCE_DIR}/src/libcxx-compat/monotonic_buffer_resource.cpp
+    ${libddwaf_SOURCE_DIR}/src/vendor/fmt/format.cc
+    ${libddwaf_SOURCE_DIR}/src/vendor/radixlib/radixlib.c
+    ${libddwaf_SOURCE_DIR}/src/vendor/lua-aho-corasick/ac_fast.cxx
+    ${libddwaf_SOURCE_DIR}/src/vendor/lua-aho-corasick/ac_slow.cxx
+    ${libddwaf_SOURCE_DIR}/src/vendor/lua-aho-corasick/ac.cxx
+    ${libddwaf_SOURCE_DIR}/src/vendor/libinjection/src/xss.c
+    ${libddwaf_SOURCE_DIR}/src/vendor/libinjection/src/libinjection_html5.c
+    ${libddwaf_SOURCE_DIR}/src/vendor/libinjection/src/libinjection_xss.c
+    ${libddwaf_SOURCE_DIR}/src/vendor/libinjection/src/libinjection_sqli.c
```
cataphract, line 129: I'm not quite sold on this. What if you need to compile the third-party projects with different flags in the future? This may also increase build times a bit because tests will not compile these files again.

CMakeLists.txt
```diff
@@ -118,6 +118,15 @@ set(LIBDDWAF_SOURCE
     ${libddwaf_SOURCE_DIR}/src/transformer/js_decode.cpp
     ${libddwaf_SOURCE_DIR}/src/transformer/common/utf8.cpp
     ${libddwaf_SOURCE_DIR}/src/libcxx-compat/monotonic_buffer_resource.cpp
+    ${libddwaf_SOURCE_DIR}/src/vendor/fmt/format.cc
```
cataphract, line 121: why not std::fmt?

Anilm3: I tried but it's not widely supported on all of our builds, support seems to be available in gcc 13, libcxx-14 (although it seems to be experimental even in 16) :
- Mac OS still uses llvm-13
- Our older docker builds use llvm-12
- MinGW builds with gcc-10

We could conditionally use one or the other but only windows compiles without adding extra libraries (e.g. libcxx-experimental for llvm-16).

CMakeLists.txt
Anilm3: > What if you need to compile the third-party projects with different flags in the future?
We've had these for ages and haven't needed to use different flags, we can always move the specific library back to `third_party` as a separate library, if that is the case.

>This may also increase build times a bit because tests will not compile these files again.
For a while now the tests use `libddwaf_objects` as a dependency so no need to recompile anything.

estringana (approved):

cataphract (approved):

PR #231, opened 2023-11-13
Vendorize re2 and utf8proc

Anilm3:
Vendorise remaining dependencies of `libddwaf.so/a`. This will eventually make it simpler (and more efficient) to support LTO and it has also allowed us to remove some unnecessary RE2 features.


estringana (approved):

cataphract (approved):
It's fine, but this tendency of putting everything in the main CMakeLists.txt is making it a bit unwieldy.

Anilm3:
> It's fine, but this tendency of putting everything in the main CMakeLists.txt is making it a bit unwieldy.

Funny that you mention that, I was thinking the same thing and refactored the main CMakeLists in https://github.com/DataDog/libddwaf/pull/232

PR #239, opened 2023-11-20
[PoC] Suspicious attacker blocking

Anilm3:
- Allow overriding rule actions through exclusion filters. With this change, rule filters can now support custom actions, in addition to `bypass` and `monitor`. These actions override specific rule actions, however custom actions in exclusion filters have a lower precedence to `bypass` and `monitor`, meaning that if two filters affect the same rule, the order of precedence is `bypass > monitor > custom action`.
- Add naive support for `rules_data` on exclusion filters


Anilm3:
Suspicious attacker blocking implemented in https://github.com/DataDog/libddwaf/pull/316 and https://github.com/DataDog/libddwaf/pull/303

PR #246, opened 2024-01-17
Fix benchmark variances

Anilm3:
- Fix some issues on valid test generation.
- Removed "mixed" tests.
- Added fixed-size tests with the nomenclature: `random.<container depth>x<container size>.<string length>


Anilm3:
The variability of benchmarks is due to the fact that tests in master are essentially more complex and buggy than the ones in this PR.

PR #247, opened 2024-01-17
Benchmark improvements

Anilm3:
- Remove random object generation
- Change tests so that the same input is evaluated on all iterations
- Allow the ability to specify common and test-specific fixtures
- Add test-specific fixtures (`eval.valid`)


Anilm3:
Benchmark is failing since the test name has been changed (so no tests are common with master). This should be resolved post-merge.

estringana (approved):

PR #248, opened 2024-01-19
Random benchmarks

Anilm3:
This PR introduces a few changes based on the new requirement to run all iterations of a scenario + fixture with the same data. The changes introduced are:

- Improved random object generation to produce an object matching the specification, albeit with a randomised structure.
- Add a new fixture for `random` tests which uses the default parameters for the object generator.

Related Jira: [APPSEC-51223]

[APPSEC-51223]: https://datadoghq.atlassian.net/browse/APPSEC-51223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


benchmark/output_formatter.cpp
```diff
@@ -5,7 +5,8 @@
 // (https://www.datadoghq.com/). Copyright 2022 Datadog, Inc.
 
 #include "output_formatter.hpp"
-#include "random.hpp"
+
+#include <cmath>
```
estringana, line 9: is this required?

estringana (approved):

benchmark/output_formatter.cpp
Anilm3: clang-18 gets very annoyed if I don't add all headers in use by the unit... but yes.

PR #251, opened 2024-02-05
SECRES-1649 - Address a libinjection false positive

Taiki-San:
libinjection interpret invalid sequences (such as words starting after a number without separators, which is causing false positives.

### DIFF
--- a/src/vendor/libinjection/src/libinjection_sqli.c
+++ b/src/vendor/libinjection/src/libinjection_sqli.c
@@ -290,18 +290,6 @@ static void st_clear(stoken_t * st)
     memset(st, 0, sizeof(stoken_t));
 }
 
-static void st_assign_char(stoken_t * st, const char stype, size_t pos, size_t len,
-                           const char value)
-{
-    /* done to eliminate unused warning */
-    (void)len;
-    st->type = (char) stype;
-    st->pos = pos;
-    st->len = 1;
-    st->val[0] = value;
-    st->val[1] = CHAR_NULL;
-}
-
 static void st_assign(stoken_t * st, const char stype,
                       size_t pos, size_t len, const char* value)
 {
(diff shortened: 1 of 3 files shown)


tests/matcher/is_sqli_test.cpp
```diff
@@ -31,12 +31,22 @@ TEST(TestIsSQLi, TestNoMatch)
...
+    };
+    
+    for (auto pattern : no_match) {
+        std::cout << pattern << std::endl;
+        ddwaf_object param;
+        ddwaf_object_string(&param, pattern);
+
+        auto match = matcher.match(param);
+        EXPECT_FALSE(match.first);
+        std::cout << match.second << std::endl;
```
Anilm3, line 46: Please remove the print statements.

Anilm3 (changes requested):
You'll need to format it using clang-17:

```
cmake .. -DCLANG_FORMAT=/usr/bin/clang-format-17
make format_fix
```

I can do it for you if you need.

tests/matcher/is_sqli_test.cpp
Taiki-San: Oops, sorry!

src/vendor/libinjection/src/libinjection_sqli.c
```diff
@@ -314,6 +302,12 @@ static void st_assign(stoken_t * st, const char stype,
     st->val[last] = CHAR_NULL;
 }
 
+static void sf_update_state(struct libinjection_sqli_state * sf, const char stype,
+                            size_t pos, size_t len, const char* value) {
+    st_assign(sf->current, stype, pos, len, value);
+    sf->hadWordBoundary = stype == TYPE_COMMENT;
```
Anilm3, line 308: `st_assign` uses `memcpy` while `st_assign_char` only had assignments. I think that's why the microbenchmark increased quite a bit:

```
scenario:is_sqli_matcher.random
🟥 execution_time [+477.773µs; +478.102µs] or [+643.703%; +644.145%]
```

Anilm3 (approved):

PR #258, opened 2024-02-10
Local file inclusion rule

Anilm3:


src/condition/lfi_detector.cpp
```diff
@@ -19,93 +17,40 @@ namespace {
...
-    std::size_t start = 0;
-    while (start < str.size()) {
-        const std::size_t end = str.find(sep, start);
-
-        if (end == start) {
-            // Ignore zero-sized strings
-            start = end + 1;
+    object::kv_iterator it(&params, {}, objects_excluded, limits);
+    for (; it; ++it) {
+        if (deadline.expired()) {
```
Taiki-San, line 28: You're checking the clock at every iteration? That's super expensive, is it not?

src/condition/lfi_detector.cpp
```diff
@@ -19,93 +17,40 @@ namespace {
...
-}
-
-std::pair<bool, std::string> lfi_impl(std::string_view path, const ddwaf_object &params)
-{
-    auto matched_params = match_params(path, params);
-    if (matched_params.empty()) {
-        return {};
-    }
+        std::string_view value{param.stringValue, static_cast<std::size_t>(param.nbEntries)};
+        if (path.find(value) == std::string_view::npos) {
```
Taiki-San, line 39: We might be able to eliminate more strings before running this (slightly expensive) check. For instance, requiring a `/` (or `\\`) to be present in the user parameter

src/condition/lfi_detector.cpp
```diff
@@ -19,93 +17,40 @@ namespace {
...
 
-    for (const auto &param : matched_params) {
-        if (param[0] == '/' && param == path) {
-            return {true, std::string(param)};
+        if (value[0] == '/' && value == path) {
+            return {{std::string(value), it.get_current_path()}};
         }
 
-        if (!path.ends_with(param)) {
+        if (!path.ends_with(value)) {
```
Taiki-San, line 47: This one is a bit scary, but trying to remove this (for instance, if `value` is a relative path and contains `../`) would almost certainly yield false positives. Having a way to optionally tweak this condition would be valuable.

src/condition/lfi_detector.cpp
Anilm3: No, the clock has a counter and only updates every 16 calls or so.

src/condition/lfi_detector.cpp
Anilm3: As discussed this isn't an issue.

src/condition/lfi_detector.cpp
Anilm3: I've added some extra checks.

estringana (approved):

src/condition/lfi_detector.cpp
```diff
@@ -19,123 +17,121 @@ namespace {
...
+    const exclusion::object_set_ref &objects_excluded, const object_limits &limits,
+    ddwaf::timer &deadline)
 {
-    auto matched_params = match_params(path, params);
-    if (matched_params.empty()) {
-        return {};
+    auto rpath = path;
+    auto drive_end = path.find(':');
+    if (drive_end != std::string_view::npos) {
+        rpath = path.substr(drive_end + 1, path.size() - (drive_end + 1));
```
Taiki-San, line 63: what would happen if the path ends with `:`?

Taiki-San (approved):

src/condition/lfi_detector.cpp
Anilm3: I wouldn't worry too much, we'll profile this later but the calls are quite fast thanks to vdso.

src/condition/lfi_detector.cpp
Anilm3: I made some changes, basically if the path ends in `:`, `rpath` will be empty so it's irrelevant for further checks. However if the parameter is the same drive (e.g. `C:`) it'll match.

Taiki-San (approved):

src/condition/lfi_detector.cpp
```diff
@@ -99,9 +102,9 @@ lfi_result lfi_impl(std::string_view path, const ddwaf_object &params,
     const exclusion::object_set_ref &objects_excluded, const object_limits &limits,
     ddwaf::timer &deadline)
 {
-    if (path.find('/') == std::string_view::npos) {
+    if (path.find('/') == npos) {
```
cataphract, line 105: the comment in the next line seems to confuse `/` used in file names and `/` used in path separators

src/condition/lfi_detector.cpp
```diff
@@ -15,14 +15,16 @@ namespace ddwaf {
 
 namespace {
 
-constexpr std::size_t min_str_len = 5;
+constexpr const auto &npos = std::string_view::npos;
```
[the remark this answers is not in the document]
Anilm3: I know but clang-tidy was being annoying so I just accepted its suggestion.

src/condition/lfi_detector.cpp
```diff
@@ -53,17 +55,19 @@ lfi_result lfi_impl_unix(std::string_view path, const ddwaf_object &params,
     return {};
```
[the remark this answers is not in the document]
Anilm3: Good point, I'll change that.

src/condition/lfi_detector.cpp
```diff
@@ -77,12 +81,11 @@ lfi_result lfi_impl_windows(std::string_view path, const ddwaf_object &params,
...
 
         std::string_view value{param.stringValue, static_cast<std::size_t>(param.nbEntries)};
-        if (value.find('\\') == std::string_view::npos || !path.ends_with(value)) {
+        if (!path.ends_with(value)) {
             continue;
         }
 
-        if ((value[0] == '\\' && value == rpath) ||
-            (value.find(':') != std::string_view::npos && value == path)) {
+        if ((value[0] == '\\' && value == rpath) || (value[1] == ':' && value == path)) {
```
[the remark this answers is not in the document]
Anilm3: You're right, I'll have to update the code to handle both cases.

src/condition/lfi_detector.cpp
```diff
@@ -36,7 +38,7 @@ lfi_result lfi_impl_unix(std::string_view path, const ddwaf_object &params,
         }
 
         std::string_view value{param.stringValue, static_cast<std::size_t>(param.nbEntries)};
-        if (value.find('/') == std::string_view::npos || !path.ends_with(value)) {
+        if (value.find('/') == npos || !path.ends_with(value)) {
```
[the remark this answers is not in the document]
Anilm3: I thought so too, but the original heuristic didn't and when you think about it, seems unlikely that an injection in the middle could end up in an exploit.

src/condition/lfi_detector.cpp
```diff
@@ -17,125 +16,116 @@ namespace ddwaf {
...
+
+    if (param.size() < min_str_len && !path.ends_with(param)) {
+        return false;
     }
 
-    for (const auto &param : matched_params) {
-        if (param[0] == '/' && param == path) {
-            return {true, std::string(param)};
+    bool is_absolute =
+        param[0] == '/' || param[0] == '\\' || (ddwaf::isalpha(param[0]) && param[1] == ':');
```
cataphract, line 67: Should be `bool is_absolute =
        param[0] == '/' || param[0] == '\\' || param.size() >= 3 &&(ddwaf::isalpha(param[0]) && param[1] == ':' && (param[2] == '/' || param[2] == '\\')));

cataphract (approved):

PR #263, opened 2024-02-13
Add weak ceilf symbol and definition

Anilm3:
Add weak `ceilf` symbol for systems in which `ceilf` isn't readily available.


CMakeLists.txt
```diff
@@ -68,8 +68,7 @@ if(LINUX)
     set(LIBDDWAF_INTERFACE_LIBRARIES
         $<$<BOOL:${LIBPTHREAD}>:pthread>
         $<$<BOOL:${LIBRT}>:rt>
-        $<$<BOOL:${LIBDL}>:dl>
-        $<$<BOOL:${LIBM}>:m>)
```
cataphract, line 72: I'm not so sure about this change. On my machine, ceil is defined in libm.so.6, not libc.so.6, so the library version will possibly not be loaded. Additionally, ceil is a weak indirect function in libm:

```
$ readelf --syms /usr/lib/x86_64-linux-gnu/libm.so.6  | grep ceil
   166: 000000000003e480    37 IFUNC   WEAK   DEFAULT   16 ceilf@@GLIBC_2.2.5
   175: 0000000000019ca0    51 FUNC    WEAK   DEFAULT   16 ceill@@GLIBC_2.2.5
   190: 000000000005b6a0   310 FUNC    WEAK   DEFAULT   16 ceilf128@@GLIBC_2.26
   256: 000000000003e480    37 IFUNC   WEAK   DEFAULT   16 ceilf32@@GLIBC_2.27
   385: 000000000002bc70    37 IFUNC   WEAK   DEFAULT   16 ceilf64@@GLIBC_2.27
   511: 000000000002bc70    37 IFUNC   WEAK   DEFAULT   16 ceilf32x@@GLIBC_2.27
   602: 000000000002bc70    37 IFUNC   WEAK   DEFAULT   16 ceil@@GLIBC_2.2.5
   919: 0000000000019ca0    51 FUNC    WEAK   DEFAULT   16 ceilf64x@@GLIBC_2.27
```

It is likely that the program where libddwaf.so is loaded will have libm loaded, but even then, both being weak symbols, I'm not sure which version will be loaded.

This is to say, perhaps it's best to use `dlsym` with a fallback to the builtin function, because it will likely be faster than your copied implementation for amd64, which can't use sse 4.1's [ROUNDSS](https://www.felixcloutier.com/x86/roundss).

cataphract (approved):

PR #268, opened 2024-02-26
Server-side request forgery (SSRF) detection operator

Anilm3:
- SSRF detection operator
- URL parsing helpers
- Fuzzer tests for both URL parser and SSRF detector
- Unit tests

Related Jira: [APPSEC-51265]

[APPSEC-51265]: https://datadoghq.atlassian.net/browse/APPSEC-51265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+using namespace std::literals;
+
+namespace ddwaf {
+
+namespace {
+
+constexpr std::array<std::string_view, 10> dangerous_ips{"169.254.0.0/16", "127.0.0.1/32",
+    "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "100.64.0.0/10", "::1/128", "fc00::/7",
+    "fe80::/10", "2001:db8:1234:1a00::/56"};
```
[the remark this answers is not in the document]
Anilm3: Indeed, good point.

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,219 @@
...
+ssrf_result ssrf_impl(const uri_decomposed &uri, const ddwaf_object &params,
+    const exclusion::object_set_ref &objects_excluded, const object_limits &limits,
+    const std::unique_ptr<matcher::ip_match> &dangerous_ip_matcher,
+    const std::unordered_set<std::string_view> &authorised_scheme_set, ddwaf::timer &deadline)
+{
+    static constexpr std::size_t min_str_len = 4;
+
+    std::string_view dangerous_domain = {};
+    for (const auto domain : dangerous_domains) {
+        if (uri.authority.raw.ends_with(domain)) {
```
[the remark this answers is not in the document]
Anilm3: Ah yes this was a mistake.

src/uri_utils.cpp
```diff
@@ -0,0 +1,220 @@
...
+    while (i < uri.size() && isschemechar(uri[i])) { ++i; }
+    if (i == 0 || i >= uri.size() || uri[i] != ':') {
+        return std::nullopt;
+    }
+
+    decomposed.scheme = uri.substr(0, i);
+
+    std::size_t next_sep = ++i;
+    if (next_sep >= uri.size()) {
+        return decomposed;
```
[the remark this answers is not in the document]
Anilm3: At this stage what we have is `<scheme>:` which is valid according to what you posted:
```
      path          = path-abempty    ; begins with "/" or is empty
                    / path-absolute   ; begins with "/" but not "//"
                    / path-noscheme   ; begins with a non-colon segment
                    / path-rootless   ; begins with a segment
                    / path-empty      ; zero characters
                    
     path-empty    = 0<pchar>

File: src/uri_utils.cpp [hunk 4]
```diff
@@ -0,0 +1,220 @@
...
+                        // The host is already malformed so we can stop here;
+                        non_ip_chars = true;
+                        break;
+                    }
+                }
+                if (!end_found || non_ip_chars || i == (host_begin + 1)) {
+                    return std::nullopt;
+                }
+
+                // Valid IPv6, remove the []
```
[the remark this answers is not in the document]
Anilm3: Fair, we assume it's IPv6 because it has the correct character set and enclosing characters, however the host can't have `[`, `]`, `:`, so I thought it was a fair assumption. We could try to parse it as well if we're being strict.

src/uri_utils.cpp
```diff
@@ -0,0 +1,220 @@
...
+                    // An empty host is a malformed authority however we might still
+                    // be able to extract a port
+                    return std::nullopt;
+                }
+
+                decomposed.authority.host = uri.substr(host_begin, i - host_begin);
+                decomposed.authority.host_index = host_begin;
+            }
+        } else {
+            // An empty host is a malformed authority
```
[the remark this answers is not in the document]
Anilm3: True...

src/uri_utils.cpp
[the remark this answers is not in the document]
cataphract: you're right

src/condition/ssrf_detector.cpp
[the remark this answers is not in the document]
Anilm3: Actually I believe the `ip_match` operator already takes care of that.

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,219 @@
...
+        if (index == std::string_view::npos) {
+            // Seemingly no injection
+            continue;
+        }
+
+        // Verify if the injected value intereferes with the authority
+        if (index < uri.scheme_and_authority.size() &&
+            (index + value.size()) > (uri.scheme.size() + 3)) {
+            // Verify if the host was modified by the injected value
+            if ((index + value.size() - 1) > uri.scheme_and_authority.size()) {
```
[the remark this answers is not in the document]
Anilm3: This slightly odd condition is meant to remove certain false positives caused by injections such as `http://domain.com/` where the injection is `domain.com/`.

I can only assume this was introduced as a result of some empirical false positives rather than assumed ones.

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,219 @@
...
+        std::string_view value{param.stringValue, static_cast<std::size_t>(param.nbEntries)};
+        auto index = uri.raw.find(value);
+        if (index == std::string_view::npos) {
+            // Seemingly no injection
+            continue;
+        }
+
+        // Verify if the injected value intereferes with the authority
+        if (index < uri.scheme_and_authority.size() &&
+            (index + value.size()) > (uri.scheme.size() + 3)) {
```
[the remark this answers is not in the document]
Anilm3: I made this a bit more clear.

.clang-tidy
```diff
@@ -1,7 +1,7 @@
...
...named-parameter,-cert-dcl37-c,-cert-dcl51-cpp,-read,-cppcoreguidelines-init-variables,-cppcoreguidelines-avoid-non-const-global-variables,-altera-id-dependent-backward-branch,-performance-no-int-to-ptr,-altera-struct-pack-align,-google-readability-casting,-modernize-use-trailing-return-type,-llvmlibc-implementation-in-namespace,-llvmlibc-callee-namespace,-cppcoreguidelines-pro-bounds-pointer-arithmetic,-fuchsia-default-arguments-declarations,-fuchsia-overloaded-operator,-cppcoreguidelines-pro-type-union-access,-fuchsia-default-arguments-calls,-cppcoreguidelines-non-private-member-variables-in-classes,-misc-non-private-member-variables-in-classes,-google-readability-todo,-llvm-header-guard,-readability-function-cognitive-complexity,-readability-identifier-length,-cppcoreguidelines-owning-memory,-cert-err58-cpp,-fuchsia-statically-constructed-objects,-google-build-using-namespace,-hicpp-avoid-goto,-cppcoreguidelines-avoid-goto,-hicpp-no-array-decay,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-pro-bounds-constant-array-index,-cppcoreguidelines-avoid-magic-numbers,-readability-magic-numbers,-abseil-string-find-str-contains,-bugprone-unchecked-optional-access'
```
Anilm3, line 4: `-bugprone-unchecked-optional-access` seems to have an issue which is causing it to get stuck in an infinite loop...

https://github.com/llvm/llvm-project/issues/59492

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,222 @@
...
+using ssrf_result = std::optional<std::pair<std::string, std::vector<std::string>>>;
+
+bool detect_parameter_injection(
+    const uri_decomposed &uri, std::string_view param, std::size_t param_index)
+{
+    const auto path_end = uri.path_index + uri.path.size();
+
+    // REST parameter injection will involve introducing a / to the URL, however, such
+    // slashes are allowed after the ? or the #. Check if the slash is within the parameters.
+    if (auto slash_index = param.find('/'); slash_index != npos) {
```
estringana, line 42: I find a bit odd this approach of two statement within the conditional. I would prioritise readability over not having a variable with longer scoper

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,222 @@
...
+    "ifconfig.pro", "localhost", "fuf.me", "localtest.me", "ulh.us" // Legacy domains
+};
+
+constexpr const std::array<std::string_view, 4> authorised_schemes{"https", "http", "ftps", "ftp"};
+
+constexpr const auto &npos = std::string_view::npos;
+
+using ssrf_result = std::optional<std::pair<std::string, std::vector<std::string>>>;
+
+bool detect_parameter_injection(
```
estringana, line 35: I can see many calls to `find` which means param will be iterated over and over. I wonder if it would be better to do a controlled iteration of the string finding they key characters rather the iterating the same many times

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,222 @@
...
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+#include "condition/ssrf_detector.hpp"
+#include "exception.hpp"
+#include "iterator.hpp"
+#include "uri_utils.hpp"
+#include "utils.hpp"
+#include <iostream>
```
estringana, line 12: I may be missing something but I don't think this is needed

estringana (approved):
Left some comments but nothing big

src/condition/ssrf_detector.cpp
Anilm3: This is perfectly readable.

src/condition/ssrf_detector.cpp
Anilm3: Perhaps although if the logic is hard to follow now, that wouldn't make it any easier...

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,279 @@
...
+        if (slash_index < path_end) {
+            // The path is partially under control of the user, this might be intentional
+            // so lets check for a possible LFI
+            auto relative_dir_index = param.find("..");
+            while (relative_dir_index != npos) {
+
+                // We found '..', check if it's enclosed by '/'
+                auto dir_index = relative_dir_index + param_index;
+                if (dir_index < path_end && (dir_index + 2) < uri.raw.size() &&
+                    uri.raw[dir_index - 1] == '/' && uri.raw[dir_index + 2] == '/') {
```
cataphract, line 71: isn't \ a possibility on windows? what about %2F (and %5C for that matter), since you're comparing against the raw (undecoded) uri?

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    // The application giving total control to the user after a given point.
+    // If it starts in the path, assume it's intentional
+    //
+    //  scheme://userinfo@host:port/path?query#fragment
+    //                             ────────────────────
+    if ((uri.query.empty() && uri.fragment.empty()) ||
+        (param_index < path_end && (param_index + param.size()) == uri.raw.size())) {
```
cataphract, line 85: how does this match the comment? FWICT, the check after the disjunction is if the param starts before path_end (and then the param finishes at the end of the raw url), not if it starts in the same place the path starts in the uri

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,279 @@
...
+        return false;
+    }
+
+    // Check if the entire query string was injected
+    //
+    //  scheme://userinfo@host:port/path?query#fragment
+    //                                 <─────
+    const auto query_index = param.find('?');
+    if (uri.query_index != npos && query_index != npos &&
+        (param_index + query_index + 1) == uri.query_index) {
```
cataphract, line 95: you don't seem to check if the *entire* query string is injected, as the comment indicates. Only whether the param starts the query string

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,279 @@
...
+        // We had some cases where this was expected behavior, to make sure
+        // we check whether the next character is a '&'
+        auto after_param = uri.raw.substr(param_index + param.length());
+        if (!after_param.empty() && after_param[0] == '&') {
+            return false;
+        }
+
+        // We can also check if the there's a valid parameter in key=value form
```
cataphract, line 103: Don't get it. What's the rationale here? How did the user format the uri, and what did he expect the param to look like?

src/condition/ssrf_detector.cpp
```diff
@@ -0,0 +1,279 @@
...
+    const exclusion::object_set_ref &objects_excluded, const object_limits &limits,
+    const std::unique_ptr<matcher::ip_match> &dangerous_ip_matcher,
+    const std::unordered_set<std::string_view> &authorised_scheme_set, ddwaf::timer &deadline)
+{
+    static constexpr std::size_t min_str_len = 4;
+
+    std::string_view dangerous_domain = {};
+    for (const auto domain : dangerous_domains) {
+        if (uri.authority.host.ends_with(domain)) {
+            dangerous_domain = domain;
```
cataphract, line 147: you can add a break

cataphract (dismissed):

src/condition/ssrf_detector.cpp
Anilm3: According to the RFC the back-slash isn't allowed as a path separator, so I omitted that entirely. The use of  %2F is a bit nuanced as in theory it's not meant to be used as a path separator but rather as a way to introduce a forward-slash that doesn't clash with the reserved forward-slash. In practice some parsers might actually consider %2F as a path separator. What do you think @Taiki-San ?

src/condition/ssrf_detector.cpp
Anilm3: You're right, the comment is inaccurate.

src/condition/ssrf_detector.cpp
Anilm3: It just refers to the fact that everything after a certain point has been injected, which might actually be the intention of the developer. I can clarify it.

src/condition/ssrf_detector.cpp
Anilm3: There was a test case in the original code covering this particular condition:

```
url: http://core-goals/v1/projects/42/goals?projectId=42&
input:  /v1/projects/42/goals?
```

cataphract: I can understand the case where you have

https://example.com/foo/{param}&a=b
with param sth like bar?c=d

or https://example.com/foo/{param}

or even (more questionable)
https://example.com/foo/{param}c=d
with param = bar? or param = bar?a=b&

but this matches even if we have param = bar?a=b

the result is something that is almost certainly a injection: they key-value pair (a, bc=d).

src/condition/ssrf_detector.cpp
cataphract: In Java, for instance, you have:

```
groovy:000> new URI('https://example.com/a%2fb').path
===> /a/b
```

you'd have to use rawPath do get the undecoded version, so you can slip a path separator this way

Anilm3: As discussed before, it might make sense to simply flag and path injection containing `%2f`... cc: @Taiki-San

src/condition/ssrf_detector.cpp
Anilm3: I'm leaning towards removing this special case as I can't think of a case that isn't questionable...

src/condition/ssrf_detector.cpp
Anilm3: I added a check to flag a parameter overlapping the path and containing %2f as an injection.

src/condition/ssrf_detector.cpp
Anilm3: I finally decided to remove the check, I think the existing checks should be enough, but we can reevaluate later.

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+            decomposed.path = uri.substr(token_begin, i - token_begin);
+
+            // We're done, nothing else to parse
+            return decomposed;
+        }
+        case token_type::authority: {
+            auto token_begin = i;
+            authority_end = uri.find_first_of("/?#", i);
+            if (authority_end != npos) {
+                // The authority is empty
```
cataphract, line 185: I don't get this comment. This would only be the case if `authority_end == i`, no?

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+                ipv4_to_ipv6(parsed_ip);
+                decomposed.authority.host_ip = parsed_ip;
+            }
+
+            if (i >= uri.size()) {
+                return decomposed;
+            }
+
+            if (uri[i] == ':') {
+                token_begin = ++i;
```
cataphract, line 290: assignment has no effect

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+                expected_token = token_type::port;
+            } else {
+                token_begin = i;
```
cataphract, line 293: idem

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+            if (!end_found || i == (token_begin + 1)) {
+                return std::nullopt;
+            }
+
+            ipaddr parsed_ip{};
+            auto host = uri.substr(token_begin + 1, i - (token_begin + 1));
+            if (!parse_ipv6(host, parsed_ip)) {
+                return std::nullopt;
+            }
+            ipv4_to_ipv6(parsed_ip);
```
cataphract, line 323: You can't enclose and ipv4 address in `[]`

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+        case token_type::port: {
+            auto token_begin = i;
+            for (; i < authority_end; ++i) {
+                if (!ddwaf::isdigit(uri[i])) {
+                    return std::nullopt;
+                }
+            }
+
+            decomposed.authority.port = uri.substr(token_begin, i - token_begin);
+            if (authority_end == uri.size()) {
```
cataphract, line 353: I guess you could validate it's in the range for uint16_t

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+                return decomposed;
+            }
+
+            expected_token = lookahead_token;
+            token_begin = i;
```
cataphract, line 358: assignment has no effect

src/matcher/ip_match.cpp
```diff
@@ -52,40 +42,49 @@ ip_match::ip_match(const std::vector<std::pair<std::string_view, uint64_t>> &ip_
...
         if (node->expiration < now) {
-            return {false, {}};
+            return false;
         }
     }
 
+    return true;
+}
+
+std::pair<bool, std::string> ip_match::match_impl(std::string_view str) const
```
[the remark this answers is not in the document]
Anilm3: This is true but I'll change it in a separate PR as I'd have to update all matchers.

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+inline bool is_unreserved(char c)
+{
+    return ddwaf::isalnum(c) || c == '-' || c == '.' || c == '_' || c == '~';
+}
+
+inline bool is_subdelim(char c)
+{
+    return c == '!' || c == '$' || c == '&' || c == '\'' || c == '(' || c == ')' || c == '*' ||
+           c == '*' || c == '+' || c == ',' || c == ';' || c == '=';
```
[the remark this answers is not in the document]
Anilm3: I'm blind, this explains why clang-tidy what complaining about this line.

src/uri_utils.cpp
```diff
@@ -0,0 +1,454 @@
...
+}
+
+// TODO validate percent-encoding?
```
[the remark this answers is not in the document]
Anilm3: For now I'm not doing it, I'll remove the comment.

src/uri_utils.cpp
Anilm3: Ah no, I had structured the code differently before and forgot to remove it.

estringana (approved):
lgtm

cataphract (approved):

PR #288, opened 2024-04-26
Use SSE4.1 ceilf when available and add badges to readme

Anilm3:
- Use the improved version of `ceilf` for amd64 from https://github.com/DataDog/nginx-datadog/pull/65
- Add some badges to the readme
- Run fuzzer on schedule and on push to master

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,11 +1,15 @@
 name: Fuzz
 
 on:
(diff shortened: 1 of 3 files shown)


Anilm3:
**Note**: I will add `cpuid` checks for SSE4.1 for the first `ceilf` call.

cataphract (approved):

PR #299, opened 2024-05-25
Transformer fixes and improvements

Anilm3:
- Fixes for vectorised lowercase transformer.
- Get rid of a branch on `remove_nulls` transformer

### DIFF
--- a/src/transformer/lowercase.cpp
+++ b/src/transformer/lowercase.cpp
@@ -41,13 +41,13 @@ bool lowercase::needs_transform(std::string_view str)
 
     const char *input = str.data();
 
-    const __m128i sse_mask_lower_bound = _mm_set1_epi8('A');
-    const __m128i sse_mask_upper_bound = _mm_set1_epi8('Z');
+    const __m128i sse_mask_lower_bound = _mm_set1_epi8('A' - 1);
+    const __m128i sse_mask_upper_bound = _mm_set1_epi8('Z' + 1);
 
     const std::size_t aligned_size = str.size() & ~0xF;
 
-    __m128i cmp_result_final = _mm_setzero_si128();
-    for (std::size_t i = 0; i < aligned_size; i += 16) {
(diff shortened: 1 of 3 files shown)


src/transformer/lowercase.cpp
```diff
@@ -56,23 +56,23 @@ bool lowercase::needs_transform(std::string_view str)
...
-        cmp_result_final = _mm_or_si128(cmp_result_final, cmp_result);
+        has_uppercase =
+            _mm_movemask_epi8(_mm_cmpeq_epi8(cmp_result, _mm_setzero_si128())) != 0xFFFF;
     }
 
-    bool is_lowercase =
-        _mm_movemask_epi8(_mm_cmpeq_epi8(cmp_result_final, _mm_setzero_si128())) == 0xFFFF;
-
-    for (std::size_t i = aligned_size; i < str.size(); i++) { is_lowercase &= !isupper(input[i]); }
+    for (std::size_t i = aligned_size; i < str.size() && !has_uppercase; i++) {
```
cataphract, line 63: it's not obvious that this will actually improve performance given that you're doing more work per loop and have more branches. Or, more exactly, that doing so will improve on average (it will likely improve for very long strings that have uppercase characters in the beginning). Did you run any benchmarks?

src/transformer/remove_nulls.cpp
```diff
@@ -24,10 +24,8 @@ bool remove_nulls::transform_impl(cow_string &str)
     std::size_t write = read;
     for (; read < str.length(); ++read) {
         auto c = str.at(read);
-        if (c == 0) {
-            continue;
-        }
-        str[write++] = c;
+        str[write] = c;
+        write += !!c;
```
cataphract, line 28: interesting. this does seem to eliminate the branch, replacing it with cmp + sbb

cataphract (approved):

PR #303, opened 2024-06-06
Custom rule filter actions

Anilm3:
This PR introduces support for custom exclusion filter actions, limited to rule filters only. The filter custom action allows overriding the "blocking" (i.e. block, redirect, monitor) action of a rule if the exclusion filter is evaluated positively. For example, when the following filter matches the given IP, `rule 1` is set to `block`:

```yaml
exclusions:
  - id: 1
    rules_target:
      - rule_id: 1
    conditions:
      - operator: ip_match
        parameters:
          inputs:
            - address: http.client_ip
          list:
            - 192.168.0.1
    on_match: block
```

The precedence of rule filter actions is still `bypass > monitor > custom`.

This feature is one of the two required for suspicious attacker blocking.

Related Jira: [APPSEC-53545]

[APPSEC-53545]: https://datadoghq.atlassian.net/browse/APPSEC-53545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/event.cpp
+++ b/src/event.cpp
@@ -190,11 +190,25 @@ void serialize_and_consolidate_rule_actions(const ddwaf::rule &rule, ddwaf_objec
     ddwaf_object_array(&actions_array);
 
     if (!action_override.empty()) {
-        ddwaf_object_array_add(&actions_array, to_object(tmp, action_override));
         auto action_it = actions.mapper.find(action_override);
         if (action_it != actions.mapper.end()) {
             const auto &[type, type_str, parameters] = action_it->second;
-            add_action_to_tracker(actions, action_override, type);
+
+            // The action override must be either a blocking one or monitor
+            if (type == action_type::monitor || is_blocking_action(type)) {
+                add_action_to_tracker(actions, action_override, type);
+            } else {
+                // Clear the action override because it's not usable
+                action_override = {};
+            }
+        } else {
+            // Without a definition, the override can't be applied
+            action_override = {};
+        }
+
+        // Tha override might have been clear if no definition was found
+        if (!action_override.empty()) {
+            ddwaf_object_array_add(&actions_array, to_object(tmp, action_override));
         }
     }
 
--- a/src/event.cpp
+++ b/src/event.cpp
@@ -178,10 +178,10 @@ void serialize_empty_rule(ddwaf_object &rule_map)
 }
 
 void serialize_and_consolidate_rule_actions(const ddwaf::rule &rule, ddwaf_object &rule_map,
-    action_type action_override, action_tracker &actions, ddwaf_object &stack_id)
+    std::string_view action_override, action_tracker &actions, ddwaf_object &stack_id)
 {
     const auto &rule_actions = rule.get_actions();
-    if (rule_actions.empty()) {
+    if (rule_actions.empty() && action_override.empty()) {
         return;
     }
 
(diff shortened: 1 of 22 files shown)


estringana (approved):

src/event.cpp
```diff
@@ -178,26 +178,31 @@ void serialize_empty_rule(ddwaf_object &rule_map)
...
 
-    if (action_override == action_type::monitor) {
-        ddwaf_object_array_add(&actions_array, to_object(tmp, "monitor"));
+    if (!action_override.empty()) {
+        ddwaf_object_array_add(&actions_array, to_object(tmp, action_override));
+        auto action_it = actions.mapper.find(action_override);
+        if (action_it != actions.mapper.end()) {
+            const auto &[type, type_str, parameters] = action_it->second;
+            add_action_to_tracker(actions, action_override, type);
+        }
```
cataphract, line 198: * is it normal behavior that an unknown action is found? If not, this should log
* a little nit, but this block should do its actions in the same order as the code below  (`add_action_to_tracker`, then the unconditional `action_override`). Better if you could merge the two branches
* Is it normal this part doesn't have the `type == action_type::generate_stack` branch? If not, one more reason to merge the two branches

src/exclusion/common.hpp
```diff
@@ -35,7 +35,8 @@ struct object_set {
 
 struct rule_policy {
     filter_mode mode{filter_mode::none};
-    std::unordered_set<const ddwaf_object *> objects{};
+    std::string_view action;
+    std::unordered_set<const ddwaf_object *> objects;
```
cataphract, line 39: could have a more descriptive name

src/exclusion/common.hpp
```diff
@@ -86,38 +88,38 @@ struct context_policy {
...
 
         const auto &p_policy = p_it->second;
         const auto &e_policy = e_it->second;
-        auto mode = p_policy.mode > e_policy.mode ? p_policy.mode : e_policy.mode;
 
-        return {mode, {p_policy.objects, e_policy.objects}};
+        if (p_policy.mode > e_policy.mode) {
+            return {p_policy.mode, p_policy.action, {p_policy.objects, e_policy.objects}};
+        }
+        return {e_policy.mode, e_policy.action, {p_policy.objects, e_policy.objects}};
```
cataphract, line 110:
```c++
const auto &effective_policy = p_policy.mode > e_policy.mode ? p_policy : e_policy;
return {
            effective_policy.mode, effective_policy.action, {p_policy.objects, e_policy.objects}};
```

cataphract (approved):

src/event.cpp
Anilm3: > is it normal behavior that an unknown action is found? If not, this should log
This could be the case if an action ID doesn't have a definition, which can happen since these are provided through different configurations, so a race condition isn't unlikely.

> Is it normal this part doesn't have the type == action_type::generate_stack branch? If not, one more reason to merge the two branches
The expectation is that the exclusion filter override must be either `monitor`, `bypass` or a blocking action such as `block` or `redirect`, I can make this expectation more explicit during serialisation.

> a little nit, but this block should do its actions in the same order as the code below (add_action_to_tracker, then the unconditional action_override). Better if you could merge the two branches
Sure

src/exclusion/common.hpp
Anilm3: I'll rename it to `action_override` as it's called elsewhere.

src/event.cpp
Anilm3: I changed the logic a bit to account for actions which perhaps weren't valid (non-blocking types).  https://github.com/DataDog/libddwaf/pull/303/commits/8c7666211cececa95436523849c34fc04d6cc2ea

PR #304, opened 2024-06-07
SHA256 hash based on OpenSSL

Anilm3:
This PR introduces a class to compute the sha256 digest, the code is based on the [openssl](https://github.com/openssl/openssl/blob/140540189c67ba94188165b1144fdfb5b248bc02/crypto/sha/sha256.c#L124) implementation. At first, the objective is to use this hash for request fingerprinting purposes, however it could be used for other purposes:

- As a processor generator, in order to produce a hash of part of the request, in a similar fashion to a fingerprint although more targeted.
- As a transformer, to match obfuscated data.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -86,6 +86,8 @@ jobs:
             params: "--dialect=sqlite"
           - fuzzer: sqli_detector
             params: "--dialect=standard"
+          - fuzzer: sha256
+            params: ""
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


estringana (approved):

cataphract (approved):
Fine. As mentioned, it would interesting to see a version using instructions dedicated to it ([SHA256H/SHA256H2/SHA256SU0/SHA256SU1](https://github.com/jocover/sha256-armv8/blob/master/sha256-armv8-aarch64.S) for aarch64) and [SHA256MSG1/SHA256MSG2/SHA256RNDS2](https://github.com/noloader/SHA-Intrinsics/blob/master/sha256-x86.c) for amd64.

PR #313, opened 2024-06-18
Rule override for adding tags

Anilm3:
This PR introduces support for adding tags to a rule through rule overrides. The schema of `rules_override` has been updated to support a new `tags` field, which is a map containing the new tags which must be added to a given set of rules. The set of tags to add must not overlap with the set of default tags already available on the rule; any overlapping tag will be ignored. For example:

```json
"rules_override": [
   {
      "rules_target": [
        {
          "tags": {
            "confidence": "1"
          }
        }
      ],
      "tags": {
          "policy": "custom-policy-1"
      }
   }
]
```

Tags will only be used for reporting purposes and won't impact other overrides or exclusion filters targeting a tag set.

Related Jira: [APPSEC-11810]

[APPSEC-11810]: https://datadoghq.atlassian.net/browse/APPSEC-11810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/rule.hpp
+++ b/src/rule.hpp
@@ -43,7 +43,7 @@ public:
     rule &operator=(const rule &) = delete;
 
     rule(rule &&rhs) noexcept = default;
-    rule &operator=(rule &&rhs) noexcept = default;
+    rule &operator=(rule &&rhs) = default;
 
     virtual ~rule() = default;
 
--- a/src/rule.hpp
+++ b/src/rule.hpp
@@ -42,23 +42,8 @@ public:
     rule(const rule &) = delete;
     rule &operator=(const rule &) = delete;
 
-    rule(rule &&rhs) noexcept
-        : enabled_(rhs.enabled_), source_(rhs.source_), id_(std::move(rhs.id_)),
-          name_(std::move(rhs.name_)), tags_(std::move(rhs.tags_)), expr_(std::move(rhs.expr_)),
-          actions_(std::move(rhs.actions_))
-    {}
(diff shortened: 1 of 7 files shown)


estringana (approved):

src/rule.hpp
```diff
@@ -42,23 +42,8 @@ class rule {
...
-        source_ = rhs.source_;
-        id_ = std::move(rhs.id_);
-        name_ = std::move(rhs.name_);
-        tags_ = std::move(rhs.tags_);
-        expr_ = std::move(rhs.expr_);
-        actions_ = std::move(rhs.actions_);
-        return *this;
-    }
+    rule(rule &&rhs) noexcept = default;
+    rule &operator=(rule &&rhs) noexcept = default;
```
cataphract, line 46: No really a good practice to specify `noexcept` in defaulted constructors. This is because `noexcept` is [inferred](https://en.cppreference.com/w/cpp/language/noexcept_spec) (provided that no base class constructor, or default member initialization or some subexpression like default argument potentially throws; but if any of those is the case, then it probably shouldn't be marked `noexcept` anyway)

cataphract (approved):

src/rule.hpp
Anilm3: Yeah you're right, I'll change it.

PR #318, opened 2024-07-08
HTTP Endpoint Fingerprint Processor

Anilm3:
This PR introduces the HTTP endpoint fingerprint processor, as well as some functionality to aid in the generation of further fingerprint fragments. To include this processor, one must add the following configuration:

```yaml
id: http-endpoint-fingerprint
generator: http_endpoint_fingerprint
conditions:
  - operator: equals
    parameters:
      inputs:
        - address: waf.context.event
      type: boolean
      value: true
parameters:
  mappings:
    - method:
        - address: server.request.method
      uri_raw:
        - address: server.request.uri.raw
      body:
        - address: server.request.body
      query:
        - address: server.request.query
      output: _dd.appsec.fp.http.endpoint
evaluate: false
output: true
```

This fragment is specifically aimed at generating a unique fingerprint for an endpoint, or rather an endpoint as it was accessed by the given request. Note that the definition of the processor has a condition on `waf.context.event`. This address will be automatically generated by libddwaf, albeit it will be added in a future PR.

Related Jira: [APPSEC-53809]

[APPSEC-53809]: https://datadoghq.atlassian.net/browse/APPSEC-53809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/sha256.cpp
```diff
@@ -178,25 +193,25 @@ std::string sha256_hash::digest()
     num = 0;
     memset(p, 0, block_size);
 
-    std::array<char, 64> final_digest{0};
-    for (unsigned int nn = 0; nn < sha_digest_length; nn += 8) {
+    for (unsigned int nn = 0; nn < N; nn += 8) {
```
Taiki-San, line 196: A more readable variable name probably wouldn't hurt 😅

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,283 @@
...
+#include "sha256.hpp"
+#include "transformer/lowercase.hpp"
+
+namespace ddwaf {
+namespace {
+
+struct string_buffer {
+    explicit string_buffer(std::size_t length)
+        // NOLINTNEXTLINE(cppcoreguidelines-no-malloc,hicpp-no-malloc,cppcoreguidelines-pro-type-reinterpret-cast)
+        : buffer(reinterpret_cast<char *>(malloc(sizeof(char) * length))), length(length)
```
Taiki-San, line 17: Wouldn't this leak if `move` isn't called?

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,283 @@
...
+    generate_fragment_field(buffer, generators...);
+
+    ddwaf_object res;
+    auto [ptr, size] = buffer.move();
+    ddwaf_object_stringl_nc(&res, ptr, size);
+
+    return res;
+}
+
+// Retur true if the first argument is less than (i.e. is ordered before) the second
```
Taiki-San, line 170:
```suggestion
// Return true if the first argument is less than (i.e. is ordered before) the second
```

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+    return left.size() <= right.size();
+}
+
+void normalize_string(std::string_view key, std::string &buffer, bool trailing_separator)
+{
+    // Clear should not deallocate...
+    // TODO: verify
+    buffer.clear();
```
Taiki-San, line 188: Does it?

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,283 @@
...
+    const unary_argument<const ddwaf_object *> &query,
+    const unary_argument<const ddwaf_object *> &body, ddwaf::timer &deadline) const
+{
+    if (deadline.expired()) {
+        throw ddwaf::timeout_exception();
+    }
+
+    // Strip query parameter from raw URI
+    auto stripped_uri = uri_raw.value;
+    auto query_idx = stripped_uri.find_first_of('?');
```
Taiki-San, line 267: That doesn't feel terribly robust

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,283 @@
...
+    if (query_idx != std::string_view::npos) {
+        stripped_uri = stripped_uri.substr(0, query_idx);
```
Taiki-San, line 269: Do we have to allocate a new string? We're just changing the length of the string and we're using `std::string_view` behind the scenes a bit later

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,283 @@
...
+void key_hash_field::operator()(string_buffer &output)
+{
+    if (value.type != DDWAF_OBJ_MAP or value.nbEntries == 0) {
+        return;
+    }
+
+    std::vector<std::string_view> keys;
+    keys.reserve(value.nbEntries);
+
+    for (unsigned i = 0; i < value.nbEntries; ++i) {
```
Taiki-San, line 232: We're only doing a top-level scanning? What happens if the body has a map under a top level parameter?

Taiki-San (approved):

src/processor/fingerprint.cpp
Anilm3: Fixed.

src/processor/fingerprint.cpp
Anilm3: We did define this as only "top-level" fields, I'm not sure we should have a special case for a single map as we can't know for sure if that's always the schema of the body or just a one off.

src/processor/fingerprint.cpp
Anilm3: Should be robust enough as the `?` character isn't allowed in the [URI path](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1:~:text=zero%2Dlength%20segment%20without%20any%20colon%20%22%3A%22-,pchar,-%3D%20unreserved%20/%20pct%2Dencoded%20/%20sub%2Ddelims%20/%20%22%3A%22%20/%20%22%40%22%0A%0A%20%20%20query).

src/processor/fingerprint.cpp
Anilm3: I'll remove the comment, as far as I can see on libcxx it just sets the size to 0.

src/processor/fingerprint.cpp
Anilm3: Not really as the view doesn't own the memory and just allows us to access memory owned by some other object (in this case the original `ddwaf_object`). A substring to a view just changes the start pointer and the size, but the underlying memory is the same.

src/sha256.cpp
Anilm3: I'll rename `N` to `DigestLength`, I want to keep the rest similar to OpenSSL.

src/processor/fingerprint.cpp
Anilm3: Ah indeed, forgot about fragments entirely.

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+template <typename T, typename... Rest>
+std::size_t generate_fragment_length(T &generator, Rest... rest)
+{
+    if constexpr (sizeof...(rest) > 0) {
+        return generator.length() + 1 + generate_fragment_length(rest...);
+    } else {
+        return generator.length();
+    }
+}
```
cataphract, line 165: Simpler with folding:

```c++
template <typename... Generators>
std::size_t generate_fragment_length(Generators &... generators)
{
    static_assert(sizeof...(generators) > 0, "At least one generator is required");
    return (generators.length() + ...) + sizeof...(generators) - 1;
}
```

src/processor/fingerprint.cpp
```diff
@@ -0,0 +1,307 @@
...
+            buffer.append(R"(\,)");
+        } else {
+            buffer.append(1, ddwaf::tolower(c));
+        }
+    }
+
+    if (trailing_separator) {
+        buffer.append(1, ',');
+    }
+}
```
cataphract, line 229: these functions operating byte by byte are good candidates for some optimizations in the future

cataphract (approved):

src/parser/processor_parser.cpp
```diff
@@ -24,22 +27,26 @@ std::vector<processor_mapping> parse_processor_mappings(
     for (const auto &node : root) {
         auto mapping = static_cast<parameter::map>(node);
 
-        // TODO support n:1 mappings and key paths
-        auto inputs = at<parameter::vector>(mapping, "inputs");
-        if (inputs.empty()) {
-            throw ddwaf::parsing_error("empty processor input mapping");
-        }
+        std::vector<processor_parameter> parameters;
+        for (const auto &param : T::param_names) {
```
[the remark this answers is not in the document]
Anilm3: Fair, I'll update this.

PR #320, opened 2024-07-15
HTTP Header and Network Fingerprints

Anilm3:


src/processor/fingerprint.cpp
```diff
@@ -298,4 +517,104 @@ std::pair<ddwaf_object, object_store::attribute> http_endpoint_fingerprint::eval
     return {res, object_store::attribute::none};
 }
 
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+std::pair<ddwaf_object, object_store::attribute> http_header_fingerprint::eval_impl(
```
estringana, line 521: Is there any alternative to this function name?

estringana (approved):

src/processor/fingerprint.cpp
Anilm3: The name is based on the fact that we're using CRTP and this is the implementation of the eval function, which is only called by the base class' eval.

src/processor/fingerprint.cpp
```diff
@@ -259,14 +369,127 @@ void key_hash_field::operator()(string_buffer &output)
...
+        if ((i + 1) < value.size()) {
+            hasher << ",";
+        }
+    }
+    hasher.write_digest(output.subspan<8>());
+}
+
+void kv_hash_fields::operator()(string_buffer &output)
+{
+    if (value.nbEntries == 0) {
```
cataphract, line 402: shouldn't you check that it's indeed a map?

src/processor/fingerprint.cpp
```diff
@@ -298,4 +521,104 @@ std::pair<ddwaf_object, object_store::attribute> http_endpoint_fingerprint::eval
...
+            child.parameterName, static_cast<std::size_t>(child.parameterNameLength)};
+
+        normalize_header(header, normalized_header);
+        auto [type, index] = get_header_type_and_index(normalized_header);
+        if (type == header_type::ip_origin) {
+            ip_origin_bitset[index] = '1';
+            // Verify not only precedence but also type, as a header of an unexpected
+            // type will be unlikely to be used unless the framework has somehow
+            // broken down the header into constituent IPs
+            if (chosen_header > index && child.type == DDWAF_OBJ_STRING) {
```
cataphract, line 589: the chosen header is not necessarily the one with highest precedence. It depends on configured, if the ip is valid, if it is non-public and there other public ones

cataphract (approved):

src/processor/fingerprint.cpp
Anilm3: I'm aware but we don't have this information so I went for the standard behaviour. Maybe we can add a new address to propagate this information for the next iteration. Since this fingerprint fragment is.... not very useful yet, I think it should be okay to improve it later.

src/processor/fingerprint.cpp
Anilm3: True, let me add that check

PR #321, opened 2024-07-15
Exists operator and `waf.context.event` virtual address

Anilm3:
This PR introduces:
- The `waf.context.event` virtual address, which is introduced within the object store whenever there is an event within the context's lifetime. This address is introduced **after** the rules have been evaluated, meaning that it can only be used with post-processors.
- The `exists` operator, which asserts the existence / presence of an address.

Related Jira: [APPSEC-54125]

[APPSEC-54125]: https://datadoghq.atlassian.net/browse/APPSEC-54125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/condition/exists.hpp
```diff
@@ -0,0 +1,35 @@
+// Unless explicitly stated otherwise all files in this repository are
+// dual-licensed under the Apache-2.0 License or BSD-3-Clause License.
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+#pragma once
+
+#include "condition/structured_condition.hpp"
+#include "matcher/ip_match.hpp"
```
cataphract, line 10: unintended include

cataphract (approved):

PR #323, opened 2024-07-17
Fingerprint fuzzer

Anilm3:
Fuzzers to validate fingerprints, currently each fuzzer generates suitable data for each fingerprint type, including known headers. As an improvement, random object types could be included on maps.


Anilm3:
Merging back to https://github.com/DataDog/libddwaf/pull/320

PR #324, opened 2024-07-18
Add function to obtain available actions

Anilm3:
This PR adds a function to report the available action types, based on what can be triggered based on the current set of rules and exclusion filters. The function in question is `ddwaf_known_actions`.

This function has been added as a request from @eliottness

### DIFF
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -247,7 +247,8 @@ void ddwaf_destroy(ddwaf_handle handle);
  *             size will be 0 if the return value is NULL.
  * @return NULL if empty, otherwise a pointer to an array with size elements.
  *
- * @Note The returned array should be considered invalid after calling ddwaf_destroy
+ * @note This function is not thread-safe
+ * @note The returned array should be considered invalid after calling ddwaf_destroy
  *       on the handle used to obtain it.
  **/
 const char* const* ddwaf_known_addresses(const ddwaf_handle handle, uint32_t *size);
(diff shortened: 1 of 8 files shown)


include/ddwaf.h
```diff
@@ -251,7 +251,23 @@ void ddwaf_destroy(ddwaf_handle handle);
...
-
+/**
+ * ddwaf_known_actions
+ *
+ * Get an array of all the action types which could be triggered as a result of
+ * the current set of rules and exclusion filters.
+ *
+ * The memory is owned by the WAF and should not be freed.
+ *
+ * @param Handle to the WAF instance.
```
estringana, line 262: Not a big deal but it should be lowercase `handle`

estringana (approved):

include/ddwaf.h
```diff
@@ -251,7 +251,23 @@ void ddwaf_destroy(ddwaf_handle handle);
...
+ * The memory is owned by the WAF and should not be freed.
+ *
+ * @param handle Handle to the WAF instance.
+ * @param size Output parameter in which the size will be returned. The value of
+ *             size will be 0 if the return value is NULL.
+ * @return NULL if empty, otherwise a pointer to an array with size elements.
+ *
+ * @Note The returned array should be considered invalid after calling ddwaf_destroy
+ *       on the handle used to obtain it.
+ **/
```
cataphract, line 269: I guess it should be noted the function is not thread-safe too.

src/ruleset.hpp
```diff
@@ -123,6 +123,36 @@ struct ruleset {
...
+            for (const auto &[name, filter] : rule_filters) {
+                auto it = actions->find(filter->get_action());
+                if (it == actions->end()) {
+                    continue;
+                }
+                available_action_types_set.emplace(it->second.type_str);
+            }
+
+            // We generate everything at the end so in the future we can change
+            // the std::unordered_set for a flat_set.
```
cataphract, line 148: I don't think you need it at all, as a class member at least. You're only using it to reclaim back the memory, but afaics it's a self-inflicted problem (you don't need to copy `type_str`)

src/ruleset.hpp
```diff
@@ -123,6 +123,36 @@ struct ruleset {
...
+            for (const auto &type : available_action_types_set) {
```
cataphract, line 149: you can reserve, as you know the size in advance

cataphract (approved):

PR #328, opened 2024-07-26
Split collections by module

Anilm3:
This PR introduces a very small change to ensure two rules with different `module` but same `type` aren't grouped in the same collection. This ensures that `type` collisions don't prevent RASP rules from being evaluated.

### DIFF
--- a/src/ruleset.hpp
+++ b/src/ruleset.hpp
@@ -26,18 +26,21 @@ struct ruleset {
     {
         rules.emplace_back(rule);
         std::string_view type = rule->get_tag("type");
-        collection_types.emplace(type);
+        std::string_view mod = rule->get_tag_or("module", "waf");
+
+        auto [it, res] = collection_types.emplace(ddwaf::fmt::format("{}.{}", mod, type));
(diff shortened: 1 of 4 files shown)


src/ruleset.hpp
```diff
@@ -169,7 +172,7 @@ struct ruleset {
     std::shared_ptr<action_mapper> actions;
 
     // The key used to organise collections is rule.type
-    std::unordered_set<std::string_view> collection_types;
+    std::unordered_set<std::string> collection_types;
```
cataphract, line 175: I'd have a slight preference see some struct along the lines of

```c++
struct collection_key {
    std::string_view type;
    std::string_view module;
}
```

with operator== and hash function, but if you're going this route you should update the comment above.

cataphract (approved):

src/ruleset.hpp
Anilm3: As mentioned offline, this is a fair point but being a quick fix and not something I plan to keep for long, I'll just update the comment.

PR #330, opened 2024-08-02
Fix false positive on SQLi EOL comments

Anilm3:
If an injection is within a comment, it can be matched as a false positive if the comment is not the first or last token in the resource.  To avoid these false positives, we must ensure that the comment itself was part of the injection and / or more than one token was injected.

Related Jiras: [APPSEC-54410]

[APPSEC-54410]: https://datadoghq.atlassian.net/browse/APPSEC-54410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/condition/sqli_detector.cpp
+++ b/src/condition/sqli_detector.cpp
@@ -96,16 +96,15 @@ bool is_query_comment(const std::vector<sql_token> &resource_tokens,
      *                              \---/
      *                            injected string
      */
-    bool is_not_last_token = param_tokens_begin < (resource_tokens.size() - 1);
     for (std::size_t i = 0; i < param_tokens.size(); ++i) {
         if (param_tokens[i].type == sql_token_type::eol_comment) {
-            if (i == 0 && param_tokens.size() == 1 && is_not_last_token) {
+            if (param_tokens.size() == 1 && param_tokens_begin < (resource_tokens.size() - 1)) {
                 // If the first and only token is the comment, ensure that it was introduced
                 // by the injection itself, rather than it being a partial match
                 return param_index <= resource_tokens[param_tokens_begin].index;
             }
 
-            return i > 0 || is_not_last_token;
+            return i > 0;
         }
     }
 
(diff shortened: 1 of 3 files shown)


estringana (approved):

src/condition/sqli_detector.cpp
```diff
@@ -95,9 +96,16 @@ bool is_query_comment(const std::vector<sql_token> &resource_tokens,
...
     for (std::size_t i = 0; i < param_tokens.size(); ++i) {
         if (param_tokens[i].type == sql_token_type::eol_comment) {
-            return i > 0 || param_tokens_begin < (resource_tokens.size() - 1);
+            if (i == 0 && param_tokens.size() == 1 && is_not_last_token) {
+                // If the first and only token is the comment, ensure that it was introduced
+                // by the injection itself, rather than it being a partial match
+                return param_index <= resource_tokens[param_tokens_begin].index;
+            }
+
+            return i > 0 || is_not_last_token;
```
Taiki-San, line 108:
```suggestion
            return i > 0;
```

src/condition/sqli_detector.cpp
```diff
@@ -95,9 +96,16 @@ bool is_query_comment(const std::vector<sql_token> &resource_tokens,
      *                              \---/
      *                            injected string
      */
+    bool is_not_last_token = param_tokens_begin < (resource_tokens.size() - 1);
     for (std::size_t i = 0; i < param_tokens.size(); ++i) {
         if (param_tokens[i].type == sql_token_type::eol_comment) {
-            return i > 0 || param_tokens_begin < (resource_tokens.size() - 1);
+            if (i == 0 && param_tokens.size() == 1 && is_not_last_token) {
```
Taiki-San, line 102:
```suggestion
            if (i == 0 && param_tokens.size() == 1) {
```

src/condition/sqli_detector.cpp
```diff
@@ -95,9 +96,16 @@ bool is_query_comment(const std::vector<sql_token> &resource_tokens,
      *                              \---/
      *                            injected string
      */
+    bool is_not_last_token = param_tokens_begin < (resource_tokens.size() - 1);
```
Taiki-San, line 99:
```suggestion
```

Taiki-San (commented):
I don't see any case where `is_not_last_token` is useful. Unless the test suite complain, I think it may be a Sqreen bug

Taiki-San (approved):

PR #333, opened 2024-08-14
Shell injection for array-based resources

Anilm3:
As an extension to the current shell injection detection operator, this PR introduces support for detecting injections on arrays of arguments, as would typically be provided to an exec-like call, albeit with an underlying shell executing said arguments rather than a direct or indirect invocation of the exec system call, which wouldn't involve any evaluation of shell code.

The substring search for arrays is slightly different:
- The provided arguments are merged to generate a single string resource.
- An abstraction (`shell_argument_array`) keeps track of both the composed string resource as well as the range of each argument, it also provides an interface compatible with `match_iterator`.
- When the `find` method is used on the `shell_argument_array`:
  - First a substring search is performed on the composed resource.
  - If the substring is found, the `[start, end]` indices are checked to verify that the substring match is within a single argument and not overlapping multiple arguments.
  - If the substring is within a single argument, the `start` index is returned, otherwise another search is performed starting `start + 1`.
  - If no valid substrings are found, `npos` is returned.

The search algorithm is meant to minimize cache misses by performing each search on consecutive blocks of memory (contiguous string buffer, contiguous indices array). Other than that, the rest of the heuristic is equivalent to the one used for strings.

Related Jira: [APPSEC-54534]

[APPSEC-54534]: https://datadoghq.atlassian.net/browse/APPSEC-54534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/condition/shi_detector.cpp
```diff
@@ -34,7 +34,105 @@ struct shi_result {
...
+            resource.append(str);
+        }
+    }
+
+    std::size_t find(std::string_view str, std::size_t start = 0)
+    {
+        while ((start = resource.find(str, start)) != npos) {
+            auto end = start + str.size() - 1;
+
+            // Ensure that both start and end are within the same argument
```
Taiki-San, line 91: Why run this algorithm on the concatenated string rather than the array from the beginning ?

Taiki-San (approved):

estringana (approved):

src/condition/shi_detector.cpp
Anilm3: The idea was that searching within a single string might have better cache properties than iterating and searching within multiple different strings, which would result in multiple indirections (array + string). Although I have not benchmarked both options.

cataphract (approved):

src/condition/shi_detector.cpp
```diff
@@ -34,7 +34,105 @@ struct shi_result {
...
+                }
+
+                // If the start and end are within the boundaries of this
+                // argument, we have determined that there is no overlap
+                if (arg_start <= start && arg_end >= end) {
+                    return start;
+                }
+
+                // Otherwise, there's overlap and it's not a valid match.
+                break;
```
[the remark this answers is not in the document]
Anilm3: This seems as though it should work but it doesn't...  I thought maybe the check should be `start_pair.first < cur.first` so that false implies `start_pair.first >= cur.first`, but even then it doesn't work because it does a binary search and it needs to check in the other direction (`iter < value`).

This works:

```cpp
            auto end = start + str.size() - 1;
            // Lower bound returns the first element where the condition is false,
            // which must be equivalent to cur < start_pair for the binary search to
            // work as expected.
            auto res = std::lower_bound(indices.begin(), indices.end(), std::pair{start, 0},
                [](const auto &cur, const auto &start_pair) {
                    return cur.second < start_pair.first;
                });

            if (res != indices.end() && res->first <= start && res->second >= end) {
                return start;
            }
            // Otherwise, there's overlap and it's not a valid match.
```

It is somewhat harder to reason though.

PR #334, opened 2024-08-15
Extend exists operator to support key paths and negation

Anilm3:
This PR adds support for asserting the presence of key paths when using the exists operator, such as through the following example:

```yaml
  - id: 2
    name: rule2-exists-kp
    tags:
      type: flow
      category: category
    conditions:
      - operator: exists
        parameters:
          inputs:
            - address: input-2
              key_path: ["path"]
```

In addition, the `!exists` operator has also been introduced. Since addresses are provided in stages, it's not currently possible to safely assert the absence of an address, hence the operator can only be used with key paths, and limited to a single address, for example:

```yaml
  - id: 2
    name: rule2-not-exists-kp
    tags:
      type: flow
      category: category
    conditions:
      - operator: "!exists"
        parameters:
          inputs:
            - address: input-2
              key_path: ["path"]
```


src/condition/exists.cpp
```diff
@@ -0,0 +1,123 @@
...
+search_outcome exists(const ddwaf_object *root, std::span<const std::string> key_path,
+    const exclusion::object_set_ref &objects_excluded, const object_limits &limits)
+{
+    if (key_path.empty()) {
+        return search_outcome::found;
+    }
+
+    // Since there's a key path, the object must be a map
+    if (root->type != DDWAF_OBJ_MAP) {
+        return search_outcome::not_found;
```
Taiki-San, line 39: Maybe unknown would make more sense?

Taiki-San (approved):

estringana (approved):

src/condition/exists.cpp
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            if (child.type != DDWAF_OBJ_MAP) {
+                return search_outcome::not_found;
+            }
+
+            // Reset the loop and iterate child
+            parent = &child;
+            i = 0;
+            size = std::min(static_cast<uint32_t>(child.nbEntries), limits.max_container_size);
```
cataphract, line 76: a bit awkward, it would be better separating the map key find from the span+object graph traversal.

cataphract (approved):
Looks good, you can mayb

PR #341, opened 2024-10-01
Object view: read only abstraction to ddwaf_object

Anilm3:
This PR introduces the `object_view` class which is a zero-cost abstraction on top of `ddwaf_object` with the objective of:
- Providing a more convenient interface to access the provided data.
- Adding generalised debug assertions which apply to every access and use of objects.
- Abstracting away the internals of how a `ddwaf_object` is structured, paving the way for the changes introduced in https://github.com/DataDog/libddwaf/pull/293.

In addition, further guarantees and checks can be added to each access, for now the interface mimics the use of an `std::optional`, where a check must be performed before an access. This applies to accessing fields (nullability check), accessing elements (size check) or accessing the contained values (type check).


src/condition/cmdi_detector.cpp
```diff
@@ -382,20 +375,20 @@ std::optional<shi_result> cmdi_impl(const ddwaf_object &exec_args,
...
+    kv_iterator it(params, {}, objects_excluded, limits);
     for (; it; ++it) {
         if (deadline.expired()) {
             throw ddwaf::timeout_exception();
         }
 
-        const ddwaf_object &param = *(*it);
-        if (param.type != DDWAF_OBJ_STRING) {
+        const object_view param = *it;
+        if (!param.is<std::string_view>()) {
```
cataphract, line 385: Why not just is_string, is_unsigned, is_signed, is_bool? I mean, this doesn't read right. param is is not a string_view

src/mkmap.hpp
```diff
@@ -16,21 +16,19 @@
 
 namespace ddwaf {
 template <typename Key, typename T, class Compare = std::less<Key>,
-    typename = std::enable_if_t<std::is_copy_constructible_v<std::remove_cv_t<std::decay_t<T>>>>>
+    typename = std::enable_if_t<std::is_copy_constructible_v<std::decay_t<T>>>>
```
cataphract, line 19: might as well modernize for c++20 here too, like below

cataphract (approved):
Nt much to say besides the reservations/suggestions I already transmitted on slack (null safety/type safety/better abstraction for keyed values). I also find it a bit template-happy in certain places, which hurts readability (esp. when the `template` keyword is needed on call sites) and makes for worse discover/error messages.

PR #349, opened 2024-10-21
Fix build on macos-14

Anilm3:
Due to missing `memory_resource` in the OSX deployment target `12` when using xcode 15 (minimum target being OSX 14). This PR uses xcode 14 which should instead build with the experimental memory resource headers and symbols.

Note that xcode 15 has deprecated the experimental memory resource.

### DIFF
--- a/cmake/try_mem_resource_ftm.cpp
+++ b/cmake/try_mem_resource_ftm.cpp
@@ -1,0 +1,8 @@
+#include <memory_resource>
+#include <version>
+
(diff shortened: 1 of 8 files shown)


estringana (approved):

cmake/try_mem_resource_ftm.cpp
```diff
@@ -0,0 +1,8 @@
+#include <memory_resource>
```
Anilm3, line 1: I kept these new checks as the diagnostics are useful.

PR #351, opened 2024-10-23
Reorganise tests 

Anilm3:
This PR reorganises unit tests into integration and unit tests, with the following definitions:
- Unit tests are those testing internals directly.
- Integration tests, as they are called at the moment, are those using the API to validate inputs and outputs.

This is the first in a number of changes that will be done around testing:
- The validator and toolswill also be folded into this directory, sharing `common` where relevant.
- The integration tests and unit tests will be split in two binaries, as this might allow testing while internals are changing.
- The fuzzer and benchmarks might be moved into this directory as well later on, unclear as of yet.


estringana (approved):
LGTM

PR #353, opened 2024-10-28
Module-based rule evaluation precedence

Anilm3:
This PR introduces the new concept of `modules` to better organise rules based on their precedence. Modules are introduced as a generic mechanism which contains a set of rules in the required order (user / base, blocking, etc). Individual modules can have optional grouping, which is introduced specifically for the purpose of having independent collections within the `waf` module, as the short-circuit evaluation follows a different criteria due to rules being grouped by type.

The modules introduced are the following: `network-acl`, `authentication-acl`, `custom-acl`, `configuration`, `business-logic`, `rasp`, `waf`. The network and authentication modules do not follow the provided timeout, and the ordering of each module changes a little bit based on whether user (Custom) or base (DD) rules should take precedence.

Related Jira: [APPSEC-55598]

[APPSEC-55598]: https://datadoghq.atlassian.net/browse/APPSEC-55598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/target_address.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace ddwaf {
+
+using target_index = std::size_t;
+
+inline target_index get_target_index(std::string_view address)
+{
+    return std::hash<std::string_view>{}(address);
+}
+
+struct target_address {
```
Anilm3, line 26: These are not currently used, but I plan to start using them soon.

estringana (approved):

Leiyks (approved):
lgtm

src/rule.hpp
```diff
@@ -16,36 +15,48 @@
...
+#include "module_category.hpp"
 #include "object_store.hpp"
 
 namespace ddwaf {
 
-class rule {
+// A core rule constitutes the most important type of entity within the
+// evaluation process. These rules are "request-bound", i.e. they are used to
+// specifically analyse request data, as opposed to other types of rules such
+// as threshold rules which analyse data across requests.
```
cataphract, line 27: Do these threshold rules actually exist?

src/rule.hpp
```diff
@@ -98,24 +112,36 @@ class rule {
...
     bool enabled_{true};
     source_type source_;
+    verdict_type verdict_{verdict_type::monitor};
     std::string id_;
     std::string name_;
     std::unordered_map<std::string, std::string> tags_;
+    std::vector<std::string> actions_;
+
+    // Frequently accessed tags
+    std::string type_;
```
cataphract, line 137: can be a string_view afaict

src/ruleset.hpp
```diff
@@ -22,33 +23,22 @@
 namespace ddwaf {
 
 struct ruleset {
-    void insert_rule(const std::shared_ptr<rule> &rule)
+    // OLINTNEXTLINE(bugprone-easily-swappable-parameters)
```
cataphract, line 26: typo

src/ruleset.hpp
```diff
@@ -22,33 +23,22 @@
...
+    void insert_rules(const std::vector<std::shared_ptr<core_rule>> &base,
```
cataphract, line 27: Do you need shared ownership so that ruleset_builder can reuse the rules in updates that don't replace the rules? Did you ensure that rules that are mutated in an update (through toggle/set_actions/set_verdict) are not incorrectly shared between rulesets?

src/builder/module_builder.hpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+inline bool base_rule_precedence(
+    const core_rule::source_type left, const core_rule::source_type right)
+{
+    return left < right;
+}
+
+inline std::string_view type_grouping_key(const core_rule *rule) { return rule->get_type(); }
+inline constexpr std::string_view null_grouping_key(const core_rule * /*rule*/) { return {}; }
```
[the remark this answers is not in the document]
Anilm3: Ah yes, I mechanically added this...

src/context.hpp
```diff
@@ -53,7 +57,7 @@ class context {
     std::vector<event> eval_rules(const exclusion::context_policy &policy, ddwaf::timer &deadline);
 
 protected:
-    bool is_first_run() const { return collection_cache_.empty(); }
+    bool is_first_run() const { return store_.empty(); }
```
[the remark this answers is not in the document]
Anilm3: Yes, you're right, I need a different indicator of first run...

src/module.cpp
```diff
@@ -0,0 +1,167 @@
...
+            if (event.has_value()) {
+                events.emplace_back(std::move(*event));
+                DDWAF_DEBUG("Found event on rule {}", rule.get_id());
+                final_verdict = verdict;
+                if (final_verdict == verdict_type::block) {
+                    break;
+                }
+            }
+        }
+        return final_verdict;
```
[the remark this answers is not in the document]
Anilm3: Yes because the `null_grouping_key` results in no collections, unless I missed something (I'll double check, but the tests wouldn't pass otherwise).

A collection only allows a single match, regardless of verdict. The module allows for multiple matches in monitor mode, so they are not exactly equivalent. Admittedly I wanted to do what you're suggesting but I then realised that new business logic rules mean that multiple rules may match at once within a given module.

src/ruleset.hpp
Anilm3: > Do you need shared ownership so that ruleset_builder can reuse the rules in updates that don't replace the rules? 
Yes, and between live handles / contexts during or after update.

> Did you ensure that rules that are mutated in an update (through toggle/set_actions/set_verdict) are not incorrectly shared between rulesets?
Currently rules are not mutated, they are regenerated when overrides are applied.

src/rule.hpp
Anilm3: Yes, in a different PR... I can dream...

src/builder/module_builder.hpp
```diff
@@ -0,0 +1,92 @@
...
+        // Configuration
+        {user_rule_precedence, null_grouping_key},
+        // Business logic
+        {user_rule_precedence, null_grouping_key},
+        // RASP
+        {base_rule_precedence, null_grouping_key},
+        // WAF
+        {user_rule_precedence, type_grouping_key},
+    }};
+};
```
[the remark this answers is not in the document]
Anilm3: Yes, albeit rules do not provide `verdict=none`. I'll update the comments with your breakdown.

cataphract (approved):

PR #354, opened 2024-10-30
Command injection detection operator 

Anilm3:
This PR introduces a new operator `cmdi_detector` which can be used to detect simple executable or shell injections in exec-style functions or methods which directly perform a fork-exec or simply exec with the given binary and arguments. The operator piggybacks on the existing shell injection heuristic to detect exploits when the injection is performed on an argument of a shell call, otherwise it's mostly used to detect executable injections. This might be expanded in the future to detect indirect injections, i.e. having more knowledge about the underlying commands. 

To use this new operator,a rule such as the following could be used:
```yaml
  - id: rsp-930-005
    name: CMDi Exploit detection
    tags:
      type: cmdi
      category: exploit_detection
      module: rasp
    conditions:
      - parameters:
          resource:
            - address: server.sys.exec.cmd
          params:
            - address: server.request.query
            - address: server.request.body
            - address: server.request.path_params
            - address: grpc.server.request.message
            - address: graphql.server.all_resolvers
            - address: graphql.server.resolver
        operator: cmdi_detector
```

In addition, this PR also adds a fuzzer for `shi_detector` (as well as one for `cmdi_detector`) with a resource of type array and extends some of the `shi_detector` tests.

Related Jira: [APPSEC-55063]

[APPSEC-55063]: https://datadoghq.atlassian.net/browse/APPSEC-55063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+std::unordered_map<std::string_view, std::string_view> known_shells{
+    {"sh", "c"},
+    {"bash", "c"},
+    {"ksh", {}},
+    {"rksh", {}},
+    {"fish", "c"},
+    {"zsh", "c"},
+    {"dash", "c"},
+    {"ash", "c"},
+    {"powershell", "Command"},
```
cataphract, line 83: looks like it would be easy to support `-CommandWithArgs` as well

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+    {"powershell.exe", "Command"},
+};
+
+std::string_view basename(std::string_view path)
+{
+    auto idx = path.find_last_of(R"(\/)"sv);
```
cataphract, line 89: this is not accurate outside of windows. `a\b` is a valid filename. So it's preferable to detect if we're being compiled for windows.

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+{
+    static const std::string_view whitespaces = " \f\n\r\t\v";
+
+    if (str.empty()) {
+        return str;
+    }
+
+    auto start = str.find_first_not_of(whitespaces);
+    if (start == std::string_view::npos) {
+        return str;
```
cataphract, line 103: you should probably return an empty string here, no?

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+std::string_view find_shell_command(std::string_view executable, const ddwaf_object &exec_args)
+{
+    // By initialising the cow_string with the underlying data
+    // contained within the std::string, we ensure a new one won't be
+    // allocated once the string is modified.
+    cow_string executable_lc{executable};
+    if (transformer::lowercase::transform(executable_lc)) {
+        executable = static_cast<std::string_view>(executable_lc);
+    }
```
cataphract, line 129: This transformation should be limited to windows.

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    auto shell_it = known_shells.find(basename(executable));
+    if (shell_it != known_shells.end()) {
+        // We've found that the current exec command is attempting to run a
+        // a shell. The shell binary itself might be injected, but also the
+        // shell command. So we need to identify the command
+        std::size_t i = 1;
+        if (!shell_it->second.empty()) {
```
cataphract, line 137: readability: name shell_it->second

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+            // Most shells allow specifying a command with -c
+            for (; i < object_size(exec_args); ++i) {
+                auto opt = trim_whitespaces(object_at(exec_args, i));
```
cataphract, line 140: not sure you should be trimming whitespace. `bash ' -c'` or `bash '-c '` is not valid syntax. Unless it's valid for other shells, there is no point

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+                    opt.find(shell_it->second) != std::string_view::npos) {
+                    // We've found the -c option, we can now break, if it isn't found
+                    // i will reach the end of the array
+                    break;
+                }
+            }
+        }
+        for (; i < object_size(exec_args); ++i) {
+            auto arg = trim_whitespaces(object_at(exec_args, i));
+            if (arg.empty() || arg.starts_with('-') || arg.starts_with('+')) {
```
cataphract, line 151: what's with the plus? what shells support this? You're missing valid commands in powershell, at least:

```
$ pwsh -Command '+1; Write-Output "Hello" '
1
Hello
```

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        // First check if the entire executable was injected
+        std::string_view value{param.stringValue, static_cast<std::size_t>(param.nbEntries)};
+        value = trim_whitespaces(value);
+        if (executable == value) {
+            // When the full binary has been injected, we consider it an exploit
+            // although bear in mind that this can also be a vulnerable-by-design
+            // application, leading to a false positive
+            return {{std::string(value), it.get_current_path()}};
```
cataphract, line 193: seems a bit strict, I mean if the program happens to execute sh, any parameter with the value 'sh' will trigger this, right?

cataphract (approved):

src/condition/cmdi_detector.cpp
Anilm3: True, I need to split the windows / linux logic.

src/condition/cmdi_detector.cpp
Anilm3: Yeah, you're right I was overthinking this.

src/condition/cmdi_detector.cpp
Anilm3: Not sure about powershell (I guess not), but you can use `shopts` in the terminal with most linux shells, e.g. `bash -e -c 'false ; ls -l'` and the equivalent with errexit disabled `bash +e -c 'false ; ls -l'`. I have a few patterns which aren't currently working so I'll have to update a bit the way we interpret command line arguments.

src/condition/cmdi_detector.cpp
Anilm3: Yes, I  agree, I already mentioned it to @Taiki-San but he didn't seem concerned...

Taiki-San: I'm not definitive on this point. If we're concerned, we can keep it in reserve for the Webshell detection

src/condition/cmdi_detector.cpp
```diff
@@ -0,0 +1,263 @@
...
+    std::vector<shell_token> &resource_tokens, const ddwaf_object &params,
+    const exclusion::object_set_ref &objects_excluded, const object_limits &limits,
+    ddwaf::timer &deadline)
+{
+    const std::string_view executable = trim_whitespaces(object_at(exec_args, 0));
+    if (executable.empty()) {
+        return {};
+    }
+
+    auto shell_command = find_shell_command(executable, exec_args);
```
Taiki-San, line 331: No early bailout in case no command were found?

Taiki-San (approved):

src/condition/cmdi_detector.cpp
Anilm3: Alright, restricting this for now to paths with more than one component (i.e. `/` or `\`).

src/condition/cmdi_detector.cpp
Anilm3: No, because we can still match the executable independently from the shell command.

PR #355, opened 2024-11-07
Disable a few patterns that caused false positives

Taiki-San:
Address a few false positives we found in the wild. The three fingerprints were not producing true positives so they'll now be ignored.


Anilm3 (approved):
lgtm, ignore the macos build failure, something must have changed.

PR #359, opened 2024-11-18
Improve parsing of numbers in SQL tokenizers

Anilm3:
This PR fixes the parsing of numbers with a sign, as the incorrect interpretation of the sign can result in false positives. For example, a statement such as:
```
SELECT values FROM table WHERE value IN (-1,-2,+3,+4);
```

Where the injected value is `-1,-2,+3,+4` should not be considered a harmful injection.

Related Jira: [APPSEC-55878]

[APPSEC-55878]: https://datadoghq.atlassian.net/browse/APPSEC-55878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/condition/ssrf_detector.hpp
```diff
@@ -13,7 +13,7 @@ namespace ddwaf {
 
 class ssrf_detector : public base_impl<ssrf_detector> {
 public:
-    static constexpr unsigned version = 1;
+    static constexpr unsigned version = 2;
```
Taiki-San: We're not doing `ssrf_detector@v2`, right?

Taiki-San (approved):

src/condition/ssrf_detector.hpp
Anilm3: I was overzealous and wanted to do everything here, but I changed this before you commented.

PR #378, opened 2025-03-09
[v2] Writable objects: owned and borrowed object

Anilm3:
This PR introduces two new object types:
- `owned_object`: as the name suggests, this is an object which owns its own memory. This abstraction ensures that the memory associated with an object is freed on destruction and moved appropriately.
- `borrowed_object`: an object type which points to memory owned by a different object, this allows accessing elements from an `owned_object` without the risk of accidental double-free.

In addition, some code from `object_view` has been refactored to ensure that `owned_object` and `borrowed_object` are both writable and readable. More of `object_view` will be refactored in the next few PRs.

Some quirks on this PR:
- The public functions `ddwaf_object_*` still contain a separate implementation, these will be updated to use the new abstractions in a following PR. This code has been moved from `object.cpp` to `interface.cpp` without changes.
- `owned_object` and `borrowed_object` currently don't check for free function compatibility. While incorrect, this shouldn't be an issue  at the moment and free functions are being replaced by allocators in a subsequent PR.

Remaining work (future PRs):
- Replace all uses of `ddwaf_object` within tests with `owned_object`, `borrowed_object` and `object_view`.
- Completely refactor `object_view` and `object_converter`.
- Remove `raw_configuration` in favour of `object_view` or use `object_view` behind the scenes.
- Update `ddwaf_object_*` interface to use new types.
- Improve context method to avoid `ddwaf_object` arguments or `optional_ref`.
- Add allocator support and check for allocator compatibility on `emplace` and `emplace_back`.
- Remove all uses of `malloc` / `free`


src/object.hpp
```diff
@@ -0,0 +1,447 @@
...
+};
+
+// NOLINTNEXTLINE(fuchsia-multiple-inheritance)
+class owned_object final : public readable_object<owned_object>,
+                           public writable_object<owned_object> {
+public:
+    using size_type = decltype(detail::object::nbEntries);
+    using length_type = decltype(detail::object::nbEntries);
+
+    owned_object() = default;
```
cataphract, line 230: I'm not not this constructor without `free_fn` (defaulting to ddwaf_object_free) makes sense. Make it private maybe?

src/object.hpp
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+    explicit borrowed_object(detail::object &obj) : obj_(&obj) {}
+
+    explicit borrowed_object(owned_object &obj);
+
+    [[nodiscard]] detail::object &ref() { return *obj_; }
+    [[nodiscard]] const detail::object &ref() const { return *obj_; }
+    [[nodiscard]] detail::object *ptr() { return obj_; }
+    [[nodiscard]] const detail::object *ptr() const { return obj_; }
```
cataphract, line 214: I imagine these are to be made private in the future, right? Same for any other public exposures of default::object.

src/object.hpp
```diff
@@ -0,0 +1,447 @@
...
+};
+
+// TODO move object view and object converter header
+
+template <typename Derived> class writable_object {
+public:
+    [[nodiscard]] borrowed_object at(std::size_t idx);
+
+    borrowed_object emplace_back(owned_object &&value);
+    borrowed_object emplace(std::string_view key, owned_object &&value);
```
cataphract, line 188: These are also unsafe and I'd move them to separate types.

src/object_view.hpp
```diff
@@ -81,14 +78,18 @@ class object_key {
     const detail::object *obj_{nullptr};
 };
 
-class object_view {
+class object_view final : public readable_object<object_view> {
```
cataphract, line 81: this is basically a read-only borrowed object right? So can't you move this to object.hpp and derive this for borrowed_object?

src/scanner.hpp
```diff
@@ -30,37 +30,33 @@ class scanner {
...
-        if (!obj.has_value() && obj.type() == object_type::invalid) {
-            return false;
+        if constexpr (std::is_same_v<T, object_view>) {
+            if (!data.has_value() || data.type() == object_type::invalid) {
+                return false;
+            }
+        } else if constexpr (std::is_same_v<T, std::string_view>) {
+            if (data.empty()) {
+                return false;
+            }
```
cataphract, line 57: You could just write overloads here.

src/object.hpp
```diff
@@ -0,0 +1,447 @@
...
+
+namespace detail {
+
+using object = ddwaf_object;
+
+inline char *copy_string(const char *str, std::size_t len)
+{
+    // TODO new char[len];
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast,hicpp-no-malloc)
+    char *copy = static_cast<char *>(malloc(sizeof(char) * (len + 1)));
```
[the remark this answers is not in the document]
Anilm3: Yeah, I don't care about this code as it'll be gone in the next few PRs with the introduction of allocators.

src/object.hpp
Anilm3: Yes, that's the eventual goal, but I can't as it is.

src/object.hpp
Anilm3: You're right although in practice is makes no difference since an invalid object doesn't need to be freed.

src/object.hpp
Anilm3: Since everything is unsafe, there's no need to distinguish it...

src/object_view.hpp
Anilm3: Somewhat, deriving it from borrowed object would make it writable and I'd have to make some methods private. In addition, I think borrowed objects will need to have the allocator as well, once I introduce support, so they won't be exactly equivalent and deriving one from the other won't make sense anymore.

As with refactoring things further and moving this type to `object.hpp`, I left a TODO about it somewhere in `object.hpp` so I'll definitely be doing that at some point.

src/scanner.hpp
Anilm3: Just a matter of preference I suppose.

src/object.hpp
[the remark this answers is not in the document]
Anilm3: Fixed

src/object.hpp
```diff
@@ -0,0 +1,447 @@
...
+    if (copy == nullptr) {
+        [[unlikely]] throw std::bad_alloc();
+    }
```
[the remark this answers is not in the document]
Anilm3: Fixed, thanks for the article, interesting read. I've been wanting to experiment with PGO and BOLT as a way to get more accurate benchmarks (i.e. removing code alignment noise), although I haven't tried yet.

src/object.hpp
Anilm3: Keeping it as is for now as it has no side-effects.

src/scanner.hpp
Anilm3: Changed.

src/object.hpp
```diff
@@ -0,0 +1,447 @@
...
+
+    obj.array = new_array;
+}
+
+inline void alloc_array(object &obj)
+{
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast,hicpp-no-malloc)
+    obj.array = reinterpret_cast<object *>(malloc(8 * sizeof(object)));
```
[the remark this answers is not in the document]
Anilm3: If `8*sizeof(object)` overflows we're in trouble anyway. I've fixed the reinterpret cast though.

src/condition/scalar_condition.cpp
```diff
@@ -65,7 +64,8 @@ ResultType eval_object(Iterator &it, std::string_view address, bool ephemeral,
                 if constexpr (std::is_same_v<ResultType, bool>) {
                     return true;
                 } else {
-                    return {{{{"input"sv, object_to_string(dst), address, it.get_current_path()}},
+                    return {{{{"input"sv, object_view{dst}.convert<std::string>(), address,
```
cataphract, line 67: this stuff could be directly on `owned_object` too

src/configuration/common/raw_configuration.hpp
```diff
@@ -27,13 +27,26 @@ class raw_configuration : public ddwaf_object {
 
     raw_configuration() = default;
     // NOLINTNEXTLINE(google-explicit-constructor,hicpp-explicit-conversions)
-    raw_configuration(const ddwaf_object &arg) : _ddwaf_object() { *((ddwaf_object *)this) = arg; }
+    raw_configuration(const ddwaf_object &arg) : _ddwaf_object()
+    {
+        dynamic_cast<ddwaf_object &>(*this) = arg;
```
cataphract, line 32: dynamic_cast?? First, this could be a `static_cast` since it's an upcast; second, can't this be written just as

```
raw_configuration(const ddwaf_object &arg) : _ddwaf_object{arg} {}
```

src/configuration/common/raw_configuration.hpp
```diff
@@ -27,13 +27,26 @@ class raw_configuration : public ddwaf_object {
...
+    {
+        dynamic_cast<ddwaf_object &>(*this) = dynamic_cast<ddwaf_object &>(other);
+        dynamic_cast<ddwaf_object &>(other) = {};
+    }
+
+    raw_configuration &operator=(raw_configuration &&other) noexcept
+    {
+        dynamic_cast<ddwaf_object &>(*this) = dynamic_cast<ddwaf_object &>(other);
+        dynamic_cast<ddwaf_object &>(other) = {};
+        return *this;
```
cataphract, line 48: similar observations

src/interface.cpp
```diff
@@ -381,4 +467,370 @@ uint32_t ddwaf_builder_get_config_paths(
...
+    if (object == nullptr) {
+        return nullptr;
+    }
+
+    // UINT64_MAX is 20 char long
+    char container[UINT64_CHARS] = {0};
+
+    const auto length = static_cast<size_t>(
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
+        snprintf(container, sizeof(container), "%" PRIu64, value));
```
cataphract, line 563: you can use the more efficient `to_chars`. same in the signed version

src/interface.cpp
```diff
@@ -381,4 +467,370 @@ uint32_t ddwaf_builder_get_config_paths(
...
+bool ddwaf_object_map_add_helper(
+    ddwaf_object *map, const char *key, size_t length, ddwaf_object object)
+{
+    if (length == SIZE_MAX) {
+        DDWAF_DEBUG("invalid key length: {}", length);
+        return false;
+    }
+
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast,hicpp-no-malloc)
+    char *name = reinterpret_cast<char *>(malloc((length + 1) * sizeof(char)));
```
cataphract, line 660: could be a static_cast

src/iterator.hpp
```diff
@@ -104,10 +106,15 @@ class key_iterator : public iterator_base<key_iterator> {
...
+        static thread_local detail::object key;
         if (current_.first.empty()) {
             return {};
         }
-        return ddwaf_object_stringl_nc(&current_key_, current_.first.data(), current_.first.size());
+        return (key = {.parameterName = nullptr,
+                    .parameterNameLength = 0,
+                    .stringValue = current_.first.data(),
+                    .nbEntries = current_.first.size(),
+                    .type = DDWAF_OBJ_STRING});
```
cataphract, line 117: not pretty. I imagine this is going away with v2?

src/object_view.hpp
```diff
@@ -434,6 +320,63 @@ class object_view {
 
 static_assert(sizeof(object_view) == sizeof(void *));
 
+inline owned_object clone(object_view input)
```
cataphract, line 323: could be method on readable_object

cataphract (approved):

src/configuration/common/raw_configuration.hpp
Anilm3: Not sure what I was thinking here

src/interface.cpp
Anilm3: I'm not touching this code, just moved it over from one file to another. It'll be nuked.

src/interface.cpp
Anilm3: Same as above.

src/object_view.hpp
Anilm3: True, I'll move it in the next PR.

src/iterator.hpp
Anilm3: To be fair it was ugly before as well but yeah this will be gone once the keys are separate objects.

src/condition/scalar_condition.cpp
Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/387/commits/95ba3b1f3430ba12ec3061a583669015d0d98deb

src/object_view.hpp
Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/387

PR #382, opened 2025-03-17
[v2] Remove object limits and defer truncation to the caller

Anilm3:
This PR removes the notion of object limits and defers the truncation of inputs to the caller. This ensures that limits can be applied selectively (or not at all) without the need for further configuration.


benchmark/main.cpp
```diff
@@ -149,8 +149,8 @@ int main(int argc, char *argv[])
             auto name = spec["scenario"].as<std::string>();
             auto ruleset = spec["ruleset"].as<ddwaf_object>();
 
-            ddwaf_config cfg{{0, 0, 0}, {nullptr, nullptr}, nullptr};
-            ddwaf_handle handle = ddwaf_init(&ruleset, &cfg, nullptr);
+            ddwaf_config config{{nullptr, nullptr}, nullptr};
```
cataphract, line 152: might as well replace with just `{}`

src/utils.hpp
```diff
@@ -45,28 +45,8 @@ struct eval_result {
...
-    static constexpr uint32_t max_key_path_depth{DDWAF_MAX_CONTAINER_DEPTH};
-
-    // User provided limits
-    uint32_t max_container_depth{DDWAF_MAX_CONTAINER_DEPTH};
-    uint32_t max_container_size{DDWAF_MAX_CONTAINER_SIZE};
-    uint32_t max_string_length{DDWAF_MAX_STRING_LENGTH};
-};
-
-inline size_t find_string_cutoff(const char *str, size_t length, object_limits limits = {})
+inline size_t find_string_cutoff(const char *str, size_t length)
```
cataphract, line 48: is this still needed?

cataphract (approved):

src/utils.hpp
Anilm3: It's not, I have to convert it into documentation since the caller will have to take make sure to truncate strings correctly.

benchmark/main.cpp
Anilm3: Yeah I'll change it in the next one.

Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/387/commits/95ba3b1f3430ba12ec3061a583669015d0d98deb

src/utils.hpp
Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/387/commits/95ba3b1f3430ba12ec3061a583669015d0d98deb

PR #383, opened 2025-03-17
Benchmarks with PGO

Anilm3:
This PR introduces two scripts to build the `waf_benchmark` binary with LTO+PGO or LTO+PGO+BOLT, the latter not performing well as of yet and taking considerably longer, therefore it is currently unused but kept for future reference. The goal of adding PGO to the benchmark only is to test the hypothesis that it could reduce alignment-related noise during typical development.


benchmark/CMakeLists.txt
```diff
@@ -5,18 +5,35 @@ target_compile_options(waf_benchmark PRIVATE $<$<BOOL:${LIBDDWAF_ENABLE_LTO}>:-f
...
 
-set_target_properties(waf_benchmark PROPERTIES
-    CXX_STANDARD 20
-    CXX_STANDARD_REQUIRED YES
-    CXX_EXTENSIONS NO)
+if (LIBDDWAF_BENCHMARK_PGO_STAGE1)
+    set_target_properties(waf_benchmark PROPERTIES
+        CXX_STANDARD 20
+        CXX_STANDARD_REQUIRED YES
+        CXX_EXTENSIONS NO
```
cataphract, line 12: you can set these common properties just once

cataphract (approved):

benchmark/CMakeLists.txt
Anilm3: True, I keep mindlessly copying this everywhere

PR #384, opened 2025-03-18
Add function to get list of loaded configuration paths

Anilm3:
This PR adds a new builder function to retrieve all loaded paths, either as a complete list or filtered based on a given regex, the signature of this function is as follows:

```c
uint32_t ddwaf_builder_get_config_paths(ddwaf_builder builder, ddwaf_object *paths, const char *filter, uint32_t filter_len);
```

This function allows the builder user to determine the available set of configurations in order to determine if the default one needs to be added.

### DIFF
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -415,6 +415,27 @@ bool ddwaf_builder_remove_config(ddwaf_builder builder, const char *path, uint32
 ddwaf_handle ddwaf_builder_build_instance(ddwaf_builder builder);
 
 /**
+ * ddwaf_builder_get_config_paths
+ *
+ * Provides an array of the currently loaded paths, optionally matching the
+ * regex provided in filter. In addition, the count is provided as the return
+ * value, allowing paths to be nullptr.
+ *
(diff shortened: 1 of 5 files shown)


include/ddwaf.h
```diff
@@ -414,6 +414,27 @@ bool ddwaf_builder_remove_config(ddwaf_builder builder, const char *path, uint32
...
+ * ddwaf_builder_get_config_paths
+ *
+ * Provides an array of the currently loaded paths, optionally matching the
+ * regex provided in filter. In addition, the count is provided as the return
+ * value, allowing paths to be nullptr.
+ *
+ * @param builder Builder to perform the operation on. (nonnull)
+ * @param paths The object in which paths will be returned, as an array of
+ *        strings. If NULL, only the count is provided. (nullable)
+ * @param filter An optional string type regex to filter the provided paths. (nullable).
```
cataphract, line 427: the documentation should specify the regex is unanchored (it's tested for all positions)

cataphract (approved):

PR #387, opened 2025-03-20
[v2] Refactor and improve object types

Anilm3:
- Refactor of `object_view` and `object_converter` to allow for more common methods on `owned_object` and `borrowed_object`.
- `ddwaf_object` interface has been updated to use owned / borrowed / view where relevant.
- Added initializer lists for container creation functions, for object creation simplicity.
- The remaining comments from https://github.com/DataDog/libddwaf/pull/378 and https://github.com/DataDog/libddwaf/pull/382 have been addressed here.

Remaining work (future PRs):

- <strike>Completely refactor object_view and object_converter.</strike>
- Replace all uses of `ddwaf_object` within tests with `owned_object`, `borrowed_object` and `object_view.`
- Remove `raw_configuration` in favour of `object_view` or use `object_view` behind the scenes.
- <strike>Update `ddwaf_object_*` interface to use new types.</strike>
- Improve context methods to avoid `ddwaf_object` arguments or `optional_ref`.
- Add allocator support and check for allocator compatibility on `emplace` and `emplace_back`.
- Remove all uses of `malloc` / `free`

PR #388, opened 2025-03-21
[v2] Improved object initializers

Anilm3:


Anilm3:
Added a more compact version of this to https://github.com/DataDog/libddwaf/pull/387

PR #391, opened 2025-03-26
[v2] Remove remaining uses of ddwaf_object in `src` and `tests/unit`

Anilm3:


Anilm3:
Changes are primarily to tests, merging...

PR #400, opened 2025-04-25
JWT Decoding Processor

Anilm3:
This PR introduces a new processor for decoding JWT tokens. This new processor can be used as a preprocessor to generate an address containing the JWT token contents as a map with the following schema:

```json
{
  "header": { ... },
  "payload": { ... },
  "signature": {
    "available" : true | false 
  }
}
```

In addition, the following supporting changes have been introduced:
- The `base64_decode` transformer has been "forked" into the `base64url_decode` transformer, as used by JWT.
- The underlying JSON within the token is being parsed using rapidjson.

Related Jiras: [APPSEC-57472]

[APPSEC-57472]: https://datadoghq.atlassian.net/browse/APPSEC-57472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/transformer/base64_decode.cpp
```diff
@@ -113,4 +76,92 @@ bool base64_decode::transform_impl(cow_string &str)
...
+    for (size_t pos = 0; pos < str.length(); ++pos) {
+        if (!ddwaf::isalnum(str[pos]) && str[pos] != '-' && str[pos] != '_') {
+            // If it's not a valid base64, it must be the trailing =
+            if (str[pos] == '=') {
+                size_t equals = 0;
+                while (pos + equals < str.length() && str[pos + equals] == '=') { equals += 1; }
+
+                // The = must go to the end, and there musn't be too many
+                const size_t padding = 4 - (pos % 4);
+                if (pos + equals == str.length() && equals <= 3 && equals <= padding) {
```
cataphract, line 137: I'm not sure about this. rfc 4648 says there can 0, 1 or 2 padding characters (section 4).

Also this method is difficult to read. Split the loop, and don't use two variable to move:

```cpp
const size_t len = str.length();
size_t pos = 0;

// consume valid characters (A-Za-z0-9-_)
while (pos < len) {
    const char c = str[pos];
    const bool is_base64url_char = (ddwaf::isalnum(c) || c == '-' || c == '_');
    if (!is_base64url_char) {
        break; // not part of the alphabet. Move to the padding check
    }
    pos++;
}

// check the remaining characters for padding ('=')
size_t padding_len = 0;
while (pos < len) {
    if (str[pos] != '=') {
        return false; // invalid char
    }
    padding_len++;
    pos++;
}

// TODO: validation for the padding amount

```

src/json_utils.cpp
```diff
@@ -0,0 +1,109 @@
...
+        } else if (doc.IsArray()) {
+            ddwaf_object_array(&object);
+        } else {
+            ddwaf_object_invalid(&object);
+        }
+        return object;
+    };
+
+    rapidjson::Document doc;
+    const rapidjson::ParseResult result = doc.Parse(json.data(), json.size());
```
cataphract, line 53: You could use the streaming parser to avoid the intermediary structures, this is especially the case here because you're not even taking advantage of the fact that you know in advance the size of the containers (the realloc overhead happened already inside rapidjson code). See e.g. https://github.com/DataDog/nginx-datadog/blob/master/src/security/body_parse/body_json.cpp#L89

CMakeLists.txt
```diff
@@ -97,7 +97,9 @@ elseif(WIN32)
     list(APPEND LIBDDWAF_INTERFACE_LIBRARIES ws2_32)
 endif()
 
-set(LIBDDWAF_PRIVATE_LIBRARIES "")
+add_subdirectory(third_party EXCLUDE_FROM_ALL)
+
+set(LIBDDWAF_PRIVATE_LIBRARIES lib_rapidjson)
```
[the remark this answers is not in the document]
Anilm3: I didn't bother because rapidjson is header only.

src/transformer/base64_decode.cpp
Anilm3: I just moved this code around (didn't really check if it was correct), I'll double check.

src/json_utils.cpp
Anilm3: Good point, I'll have a look thanks

Anilm3: > because you're not even taking advantage of the fact that you know in advance the size of the containers

In fairness, this will not be an issue in v2 since the new abstractions allow for preallocation.

Anilm3: Done.

src/transformer/base64_decode.cpp
Anilm3: Fixed.

src/json_utils.cpp
```diff
@@ -64,6 +64,19 @@ class object_reader_handler
     : public rapidjson::BaseReaderHandler<rapidjson::UTF8<>, object_reader_handler> {
 public:
     object_reader_handler() { stack_.reserve(max_depth + 1); }
+    ~object_reader_handler()
+    {
+        // Cleanup
+        ddwaf_object_free(&root_);
```
cataphract, line 70: it would be better if there were an owning ddwaf_object (maybe in v2)

src/json_utils.cpp
```diff
@@ -64,6 +64,19 @@ class object_reader_handler
...
+        if (key_ != nullptr) {
```
cataphract, line 71: prefer something like `std::unique_ptr<char[], decltype(&free)>`

cataphract (approved):

src/json_utils.cpp
Anilm3: Good point.

PR #402, opened 2025-05-07
Replace ddwaf_result with ddwaf_object

Anilm3:
This PR replaces `ddwaf_result`  with a `ddwaf_object` in the `ddwaf_run` call. This result object is only filled when the result is a non-error one and it contains a map with the following contents:

```
{
  "events": [ ... ],
  "actions": { ... },
  "timeout": (true | false),
  "duration": integer,
  "attributes": { ... },
  "keep": (true | false)
}
```

Note that `derivatives` have been replaced with `attributes` and `keep` is provided but not yet supported internally.

Related Jiras: [APPSEC-57573]

[APPSEC-57573]: https://datadoghq.atlassian.net/browse/APPSEC-57573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -108,6 +108,9 @@ std::pair<DDWAF_RET_CODE, ddwaf_object> context::run(
     auto on_exit = scope_exit([this]() { this->exclusion_policy_.ephemeral.clear(); });
 
     auto *free_fn = ruleset_->free_fn;
+
+    // TODO these checks should be moved to the interface through something along the
+    // lines of ctx.insert(...) -> bool
     if (persistent.has_value() && !store_.insert(*persistent, attribute::none, free_fn)) {
         DDWAF_WARN("Illegal WAF call: parameter structure invalid!");
         return {DDWAF_ERR_INVALID_OBJECT, {}};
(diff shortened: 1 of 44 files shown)


src/context.cpp
```diff
@@ -48,51 +51,94 @@ void set_context_event_address(object_store &store)
...
 }
 
+// NOLINTBEGIN(cppcoreguidelines-avoid-const-or-ref-data-members)
+struct result_wrapper {
+    ddwaf_object &events;
+    ddwaf_object &actions;
+    ddwaf_object &duration;
+    ddwaf_object &timeout;
+    ddwaf_object &attributes;
+    ddwaf_object &keep;
```
cataphract, line 61: looks unused

src/context.cpp
```diff
@@ -48,51 +51,94 @@ void set_context_event_address(object_store &store)
...
+};
+// NOLINTEND(cppcoreguidelines-avoid-const-or-ref-data-members)
+
+result_wrapper initialise_result_object(ddwaf_object &object)
+{
+    ddwaf_object_map(&object);
+
+    bool add_res = true;
+    ddwaf_object tmp;
+    add_res &= ddwaf_object_map_addl(&object, STRL("events"), ddwaf_object_array(&tmp));
```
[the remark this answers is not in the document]
Anilm3: No, I just wanted a lazy way to get a combined failure.

src/context.cpp
Anilm3: Yes, this will be used in the next PR (trace-tagging)

estringana (approved):

PR #403, opened 2025-05-09
Support for partial event obfuscation

Anilm3:
The goal of this PR is to change how the event obfuscation process works:
- First it introduces a new value regular expression which contains a number of capture groups aimed at identifying the parts of the value which should be obfuscated.
- Second it changes the obfuscation process, when the value regex is the default one, only the matched capture groups are redacted rather than the whole value.
- Third it changes how the highlight is obfuscated, ensuring that it's only obfuscated when the relevant value is as well rather than relying on the regular expression, which may not work on a substring of the original value.

Finally, this PR also introduces a new type called `dynamic_string` which allows making changes to a string buffer which can then be directly moved into a `ddwaf_object`, preventing the need for extra copies.

Related Jiras: [APPSEC-57575]

[APPSEC-57575]: https://datadoghq.atlassian.net/browse/APPSEC-57575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/dynamic_string.hpp
+++ b/src/dynamic_string.hpp
@@ -27,30 +27,34 @@ public:
 
     explicit dynamic_string(std::size_t capacity)
     {
-        realloc_if_needed(capacity);
+        ensure_spare_capacity(capacity);
         buffer_.get()[0] = '\0';
     }
 
-    template <typename T>
-    // NOLINTNEXTLINE(google-explicit-constructor,hicpp-explicit-conversions)
-    dynamic_string(T str)
-        requires std::is_same_v<std::string_view, std::decay_t<T>> ||
-                 std::is_same_v<std::string, std::decay_t<T>>
-        : size_(str.size())
+    dynamic_string(const char *str, std::size_t size) : size_(size)
     {
-        realloc_if_needed(size_);
-        if (!str.empty()) {
-            memcpy(buffer_.get(), str.data(), size_);
+        ensure_spare_capacity(size_);
+        if (size_ != 0) {
+            memcpy(buffer_.get(), str, size_);
         }
         buffer_.get()[size_] = '\0';
     }
 
     // NOLINTNEXTLINE(google-explicit-constructor,hicpp-explicit-conversions)
-    dynamic_string(const char *str) : dynamic_string(std::string_view{str}) {}
+    dynamic_string(std::string_view str) : dynamic_string(str.data(), str.size()) {}
+
+    dynamic_string(const char *str) = delete;
+
+    template <typename T>
+    // NOLINTNEXTLINE(google-explicit-constructor,hicpp-explicit-conversions)
+    dynamic_string(T str)
+        requires std::is_constructible_v<std::string_view, T>
+        : dynamic_string(std::string_view{str})
+    {}
 
     ~dynamic_string() = default;
 
-    dynamic_string(const dynamic_string &str) : dynamic_string(std::string_view{str}) {}
+    dynamic_string(const dynamic_string &str) : dynamic_string(str.data(), str.size()) {}
 
     dynamic_string &operator=(const dynamic_string &str)
     {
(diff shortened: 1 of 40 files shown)


src/obfuscator.cpp
```diff
@@ -23,51 +27,148 @@ bool match(re2::RE2 &regex, std::string_view value)
...
+}
 
-            sp = re2::StringPiece(default_key_regex_str.data(), default_key_regex_str.size());
-            key_regex = std::make_unique<re2::RE2>(sp, options);
+template <std::size_t N>
+bool find_and_consume(
+    re2::StringPiece *input, re2::RE2 &regex, std::array<re2::StringPiece, N> &array)
+{
+    return find_and_consume_sequence(input, regex, array, std::make_index_sequence<N>{});
+}
```
cataphract, line 60: something like:

```cpp
template <std::size_t N>
bool find_and_consume(
    re2::StringPiece *input, re2::RE2 &regex, std::array<re2::StringPiece, N> &array)
{
    return std::apply(
        [input, &regex](auto&... elements) {
            return re2::RE2::FindAndConsume(input, regex, &elements...);
        },
        array
    );
}
```

avoid recursive templates

src/obfuscator.cpp
```diff
@@ -23,51 +27,148 @@ bool match(re2::RE2 &regex, std::string_view value)
...
+        }
+        return false;
+    }
+
+    dynamic_string output{value.size()};
+    re2::StringPiece input{value.data(), value.size()};
+
+    auto prev = input;
+    while (!input.empty()) {
+        std::array<re2::StringPiece, 8> matches;
```
cataphract, line 146: Don't hardcode 8 here.

src/obfuscator.cpp
```diff
@@ -23,51 +27,148 @@ bool match(re2::RE2 &regex, std::string_view value)
...
+                break;
+            }
+        }
+        prev = input;
+    }
+
+    if (!output.empty()) {
+        output.append({prev.data(), prev.size()});
+        value = output;
+    }
```
cataphract, line 169: I think this could be simplified. You can lose the third append inside the loop and consolidate it with the first append.

```
char *orig_read = value.data;
while (!input.empty()) {
// ...
output.append(read, match.data() - read);
output.append(redaction_msg);
read = match.data() + match.size();
}
output.append(read, value.data() + value.size() - read);
```

cataphract (approved):

src/dynamic_string.hpp
```diff
@@ -0,0 +1,154 @@
...
+    // NOLINTNEXTLINE(google-explicit-constructor,hicpp-explicit-conversions)
+    operator std::string() const { return {buffer_.get(), size_}; }
+
+    bool operator==(const dynamic_string &other) const
+    {
+        return size_ == other.size_ && (memcmp(buffer_.get(), other.buffer_.get(), size_) == 0);
+    }
+
+protected:
+    void realloc_if_needed(std::size_t at_least)
```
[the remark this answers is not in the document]
Anilm3: Fine, I'll change it.

src/obfuscator.cpp
Anilm3: Yeah, good point, thanks

src/dynamic_string.hpp
```diff
@@ -0,0 +1,154 @@
...
+        realloc_if_needed(capacity);
+        buffer_.get()[0] = '\0';
+    }
+
+    template <typename T>
+    // NOLINTNEXTLINE(google-explicit-constructor,hicpp-explicit-conversions)
+    dynamic_string(T str)
+        requires std::is_same_v<std::string_view, std::decay_t<T>> ||
+                 std::is_same_v<std::string, std::decay_t<T>>
+        : size_(str.size())
```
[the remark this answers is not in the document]
Anilm3: It seems the compiler doesn't like initialiser lists with `std::string` if I do this, I've made a small change to support this.

PR #404, opened 2025-05-14
Support for attribute generation from rules

Anilm3:
This PR primarily introduces support for generating attributes through rules. This is a new mechanism which can be used to tag traces with small amounts of metadata based on the request, response data, or even RASP data. In addition, this PR also introduces two more toggles for rules:
- `event`: enables or disable event generation ,
- `keep`: indicates to the caller that any trace sampling must be bypassed for this particular event or set of attributes

These new fields are introduced in a new map named `output`, as can be seen in the update rule schema below, more details can be found in the relevant RFC:

```json
    {
      "id": ...,
      "name": ...,
      "tags": { ... },
      "conditions": [...],
      "output": {
        "event": (true | false),
        "keep": (true | false),
        "attributes": { ... }
      }
    }
```

In addition to this change, some internal refactoring has been done, which explains the size of the PR, as many tests had to be updated. Note that copilot provides a summary of changes (albeit a limited one).

Related Jiras: [APPSEC-57574]

[APPSEC-57574]: https://datadoghq.atlassian.net/browse/APPSEC-57574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/attribute_collector.hpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    bool emplace(std::string_view key, const ddwaf_object &value, bool copy);
+
+    bool collect(const object_store &store, target_index input_target,
+        std::span<const std::string> input_key_path, std::string_view output);
+
+    void collect_pending(const object_store &store);
+
+    [[nodiscard]] bool has_pending_attributes() const { return !pending_.empty(); }
```
cataphract, line 71: has no callers outside tests

src/attribute_collector.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    ddwaf_object get_available_attributes()
```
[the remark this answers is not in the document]
cataphract: I'd also simplify the interface and remove all the pending stuff from it. You just insert (put) attributes, collect and fetch them (you'd pass the store to `get_available_attributes`)

src/attribute_collector.hpp
```diff
@@ -0,0 +1,100 @@
...
+    enum class collection_state : uint8_t { success, unavailable, failed };
+
+    collection_state collect_helper(const object_store &store, target_index input_target,
+        std::span<const std::string> input_key_path, std::string_view output);
+    bool emplace_helper(std::string_view key, const ddwaf_object &value, bool copy);
+
+    // The views and spans used here are owned by rules and processors, these
+    // are part of their definition and are unchanging during the lifetime of
+    // the context, therefore it's safe to use them.
+    using target_type = std::pair<target_index, std::span<const std::string>>;
```
cataphract, line 93: wouldn't hurt to use a struct or at least mention what the span is about

src/configuration/rule_parser.cpp
```diff
@@ -60,12 +64,61 @@ rule_spec parse_rule(raw_configuration::map &rule, core_rule::source_type source
...
+    if (at<bool>(output, "event", true)) {
+        flags = flags | rule_flags::generate_event;
+    }
+    if (at<bool>(output, "keep", true)) {
+        flags = flags | rule_flags::keep_outcome;
+    }
+
+    auto attr_map = at<raw_configuration::map>(output, "attributes", {});
+    std::vector<rule_attribute> attributes;
+    for (auto &[output, value_or_target] : attr_map) {
```
cataphract, line 78: don't shadow the earlier variable

src/rule.hpp
```diff
@@ -21,33 +21,116 @@
...
+enum class rule_source : uint8_t { base = 1, user = 2 };
+
+struct rule_attribute {
+    struct input_target {
+        std::string name;
+        target_index index;
+        std::vector<std::string> key_path;
+    };
+    std::variant<input_target, std::string, uint64_t, int64_t, double, bool> input;
+    std::string output;
```
cataphract, line 66: Not a fan of the naming; input/output here are nondescript and are not symmetrical. IIUC one is the attribute value or the source of the attribute value and the other the attribute key.

src/attribute_collector.cpp
```diff
@@ -0,0 +1,105 @@
...
+        if (state != collection_state::unavailable) {
+            it = pending_.erase(it);
+        } else {
+            ++it;
+        }
+    }
+}
+
+attribute_collector::collection_state attribute_collector::collect_helper(const object_store &store,
+    target_index input_target, std::span<const std::string> input_key_path, std::string_view output)
```
cataphract, line 62: maybe some logging here

cataphract (approved):

src/attribute_collector.hpp
```diff
@@ -0,0 +1,100 @@
...
+public:
+    attribute_collector() { ddwaf_object_map(&attributes_); }
+    attribute_collector(const attribute_collector &) = delete;
+    attribute_collector &operator=(const attribute_collector &) = delete;
+    attribute_collector(attribute_collector &&other) noexcept = delete;
+    attribute_collector &operator=(attribute_collector &&other) noexcept = delete;
+    ~attribute_collector() { ddwaf_object_free(&attributes_); }
+
+    // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
+    bool emplace(std::string_view key, std::string_view value)
```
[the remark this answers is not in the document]
Anilm3: I'm constructing intermediate objects because there's no choice, but in v2 this should work as intended.

src/attribute_collector.hpp
Anilm3: Yes, I only added this for testing, I'll add a comment.

src/attribute_collector.hpp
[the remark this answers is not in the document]
Anilm3: You're right that I should've added more comments, the pending attributes are necessary because the required addresses may not be available at the time the rule matches, so they're kept in pending and checked on every run to generate them as the data becomes available.

> name suggests it's not a destructive operation
Fair enough I'll rename it to `move_available_attributes` or  `get_available_attributes_and_reset`.

src/attribute_collector.cpp
Anilm3: True I should add a bit more logging all around.

src/attribute_collector.hpp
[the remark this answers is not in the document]
Anilm3: On the other hand, if you meant changing things so that collect just adds the attribute to the pending queue, I did consider that approach initially but I'm not keen on the idea of the extra allocations when they're not strictly necessary. I can't predict how this feature will be used or abused but I assume that, in most cases, the pending queue will typically remain empty / unallocated.

src/attribute_collector.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <string_view>
+#include <unordered_map>
+
+#include "object_store.hpp"
+#include "target_address.hpp"
+#include "utils.hpp"
+
+namespace ddwaf {
+
+class attribute_collector {
```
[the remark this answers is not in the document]
Anilm3: Added some documentation.

src/attribute_collector.hpp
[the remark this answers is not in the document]
Anilm3: Renamed to insert for now.

src/attribute_collector.hpp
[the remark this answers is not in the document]
Anilm3: Renamed to `get_available_attributes_and_reset`.

src/attribute_collector.hpp
Anilm3: Done.

src/attribute_collector.hpp
Anilm3: Added a comment explaining the purpose of both elements.

src/configuration/rule_parser.cpp
Anilm3: Fixed.

src/attribute_collector.cpp
Anilm3: Added some logging.

src/rule.hpp
Anilm3: Fixed.

PR #413, opened 2025-06-06
[v2] Container view types

Anilm3:
This PR introduces `array_view` and `map_view` as new container view types to better reflect the required underlying type in cases where this is explicitly needed.

The semantics of these view types do not resemble those of STL containers, instead, they follow the semantics of `object_view` itself for consistency across `detail::object` wrappers.


src/object.hpp
```diff
@@ -498,6 +524,229 @@ class object_view final : public readable_object<object_view> {
...
+    };
+
+    [[nodiscard]] iterator begin() const { return iterator{data_, data_ + size_}; }
+
+    [[nodiscard]] iterator end() const { return iterator{data_ + size_, data_ + size_}; }
+
+protected:
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
+    const detail::object *data_{nullptr};
+    uint16_t size_{0};
```
cataphract, line 609: I doubt there's a performance advantage in copying the size here.

src/object.hpp
```diff
@@ -498,6 +524,229 @@ class object_view final : public readable_object<object_view> {
...
+        size_ = o->via.array.size;
+    }
+    // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
+    array_view(const detail::object &o) : array_view(&o) {}
+    // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
+    array_view(object_view o) : array_view(o.ptr()) {}
+    // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
+    array_view(const owned_object &ow);
+    // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
+    array_view(const borrowed_object &ow);
```
cataphract, line 547: Repetitive... maybe using some concept for types convertible to detail::object *? If not `std::convertible_to` because you don't want to make it directly convertible, some other method to expose the pointer.

src/object.hpp
```diff
@@ -498,6 +524,229 @@ class object_view final : public readable_object<object_view> {
...
+            }
+
+            if (!child.has_value() || !child.is_map()) {
+                break;
+            }
+
+            root = child;
+        }
+        return {};
+    }
```
cataphract, line 701: I think this is both too complex an operation to have here and at the same time could be more useful if it allowed also array traversal.

cataphract (approved):

src/object.hpp
Anilm3: I'll address this in the next PR.

src/object.hpp
Anilm3: Maybe, I don't think I'm using it at the moment. I need to support globs and such so I might remove it and write something more elaborate.

PR #428, opened 2025-07-04
[v2] Add memory resource to owned and borrowed objects

Anilm3:
This PR expands https://github.com/DataDog/libddwaf/pull/418 to use allocators on owned and borrowed objects, ensuring that objects are allocator-compatible and that all possible allocations are performed with the correct allocator.

All relevant changes are in `object.hpp`.

Note that the size of the PR is due to ancillary changes required due to the following changes to the owned object:
- The `nullptr_t` constructor for `owned_object` has been removed as it was causing conflicts with other constructors.
- The `make_(array|map)` functions accepting initializer lists have been extracted out into `object_builder::(map|array)` as they were also causing conflicts due to the introduction of the capacity parameter to the other overloads.

These changes can be safely ignored.

Remaining work (Next PRs):
- <del>Update objects to use allocators more effectively.</del>
- <del>Check for allocator compatibility where relevant.</del>
- Propagate allocator from the context.
- Have a clear distinction between input , output and internal allocators.
- Update interface:
  - Pass input / ouput allocators to `context_init`.
  - Pass allocators to memory-allocating `ddwaf_object` functions.

### DIFF
--- a/src/object.hpp
+++ b/src/object.hpp
@@ -18,6 +18,7 @@
 #include <initializer_list>
 #include <stdexcept>
 #include <type_traits>
+#include <variant>
 
 namespace ddwaf {
 
@@ -80,8 +81,6 @@ union [[gnu::may_alias]] object {
         object_array array;
         object_map map;
     } via;
-
-    object &operator=(owned_object &&o) noexcept;
 };
 
 struct object_kv {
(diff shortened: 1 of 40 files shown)


src/object.hpp
```diff
@@ -117,88 +116,93 @@ static_assert(offsetof(object_map, ptr) == offsetof(object_array, ptr));
...
+}
+
 // NOLINTNEXTLINE(misc-no-recursion)
-inline void object_destroy(object &obj)
+inline void object_destroy(object &obj, memory::memory_resource *alloc)
 {
-    static memory::memory_resource *alloc = memory::get_default_resource();
+    if (alloc->is_equal(*memory::get_default_null_resource())) {
+        return;
+    }
```
cataphract, line 178: This special case is unnecessary, no? You won't be using the null resource in normal conditions, so you shouldn't pay the penalty. In the unlikely event that you are, the code should still work correctly.

src/object.hpp
```diff
@@ -117,88 +116,93 @@ static_assert(offsetof(object_map, ptr) == offsetof(object_array, ptr));
...
+inline char *copy_string(const char *str, SizeType length, memory::memory_resource &alloc)
+{
+    auto *copy = alloc_helper<char, uint32_t>(length, alloc);
+    memcpy(copy, str, length);
+    return copy;
+}
+
 // NOLINTNEXTLINE(misc-no-recursion)
-inline void object_destroy(object &obj)
+inline void object_destroy(object &obj, memory::memory_resource *alloc)
```
cataphract, line 174: `alloc` can't be null here; you could signal that better by taking a reference.

src/object.hpp
```diff
@@ -1172,31 +1231,102 @@ borrowed_object writable_object<Derived>::emplace(std::string_view key, T &&valu
...
-    std::initializer_list<detail::initializer::movable_object> list)
+struct movable_object {
+    // NOLINTNEXTLINE(google-explicit-constructor,hicpp-explicit-conversions)
+    template <typename T> movable_object(T &&value) : object{std::forward<T>(value)} {}
+    movable_object(const movable_object &) = delete;
+    movable_object(movable_object &&) = delete;
+    movable_object operator=(const movable_object &) = delete;
+    movable_object operator=(movable_object &&) = delete;
+    ~movable_object() = default;
+    mutable owned_object object;
```
cataphract, line 1253: this hack would merit a comment explaining why it's needed

src/object.hpp
```diff
@@ -1172,31 +1231,102 @@ borrowed_object writable_object<Derived>::emplace(std::string_view key, T &&valu
...
+
+inline owned_object array(std::initializer_list<all_types> list = {},
+    memory::memory_resource *alloc = memory::get_default_resource())
 {
-    auto container = make_array();
-    for (const auto &value : list) { container.emplace_back(std::move(value.object)); }
+    auto container = owned_object::make_array(static_cast<uint16_t>(list.size()), alloc);
+    for (const auto &value : list) {
+        if (std::holds_alternative<bool>(value)) {
+            container.emplace_back(owned_object{std::get<bool>(value)});
```
cataphract, line 1266: I'm not sure the compiler can elide the second variant type check. Maybe `get_if`?

cataphract (commented):
Just a couple comments, with the proviso that I think allocators should be used (and maybe even integrations with the standard library like the the `allocator_type` typedef and allocator-aware constructors)

src/object.hpp
Anilm3: Good point

src/object.hpp
```diff
@@ -117,88 +116,93 @@ static_assert(offsetof(object_map, ptr) == offsetof(object_array, ptr));
...
-inline std::pair<T *, SizeType> realloc_helper(T *data, SizeType size)
+inline std::pair<T *, SizeType> realloc_helper(
+    T *data, SizeType size, memory::memory_resource &alloc)
     requires(std::is_unsigned_v<SizeType> && sizeof(SizeType) <= sizeof(std::size_t))
 {
-    // Since allocators have no realloc interface, we're just using calloc
-    // as it'll be equivalent once allocators are supported
+    static_assert(maxof_v<SizeType> <= maxof_v<std::size_t> / sizeof(T));
+
+    constexpr std::size_t max_elements = maxof_v<SizeType>;
```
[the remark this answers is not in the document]
Anilm3: Addressed in #420, in https://github.com/DataDog/libddwaf/pull/420/commits/bdbf8bbc892cd9e692ef61faf6f6382312fb78e1

src/object.hpp
Anilm3: Addressed in #420, in https://github.com/DataDog/libddwaf/pull/420/commits/bdbf8bbc892cd9e692ef61faf6f6382312fb78e1

src/object.hpp
Anilm3: Addressed in #420, in https://github.com/DataDog/libddwaf/pull/420/commits/bdbf8bbc892cd9e692ef61faf6f6382312fb78e1

cataphract (approved):

PR #437, opened 2025-07-31
Accept empty and inconsequential configurations

Anilm3:
This PR modifies the behaviour of the builder to allow configurations which have no side-effects, i.e. inconsequential configurations. These are configurations which have no errors but also no loadable items, for example, a configuration may contain rules with a `max_version` below the current one, therefore it's perfectly valid but nothing will be loaded. Similarly, a configuration may be entirely empty or contain only empty sections.

While these configs don't provide any value, they also cause no harm, therefore they are now accepted instead of rejected.

In addition, this PR also removes the `null_ruleset_info` and `base_ruleset_info` classes as production deployments will always require diagnostics.

### DIFF
--- a/src/configuration/configuration_manager.hpp
+++ b/src/configuration/configuration_manager.hpp
@@ -57,6 +57,12 @@ public:
 protected:
     void remove_config(const configuration_change_spec &cfg);
 
+    // This function returns:
+    //  - true: when the configuration was fully or partially loaded, this
+    //          will also be the return value when the configuration is empty or
+    //          inconsequential, i.e. no element can be loaded due to compat
+    //  - false: when no elements within the configuration could be loaded due
+    //           to errors. This implies that the configuration must be rejected
     static bool load(
         raw_configuration::map &root, configuration_collector &collector, ruleset_info &info);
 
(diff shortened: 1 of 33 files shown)


src/configuration/configuration_manager.hpp
```diff
@@ -57,8 +57,8 @@ class configuration_manager {
 protected:
     void remove_config(const configuration_change_spec &cfg);
 
-    static void load(
-        raw_configuration::map &root, configuration_collector &collector, base_ruleset_info &info);
+    static bool load(
```
cataphract, line 60: document what the return value means. It's not clear it's used to signal errors or more generally whether something was loaded (from the impl it seems it returns false if there are errors only).

src/ruleset_info.hpp
```diff
@@ -240,7 +182,28 @@ class ruleset_info : public base_ruleset_info {
...
-    void set_error(std::string error) override { error_ = std::move(error); }
+    void set_error(std::string error) { error_ = std::move(error); }
+
+    [[nodiscard]] ruleset_info_state state() const noexcept
+    {
+        if (!error_.empty()) {
+            return ruleset_info_state::invalid;
+        }
+
+        auto final_state = ruleset_info_state::indeterminate;
```
cataphract, line 193: not sure indeterminate passes the right message. It seems like the object is an indeterminate state, but this seems like a valid state. Maybe "empty"?

cataphract (approved):

PR #439, opened 2025-08-15
URI parsing preprocessor

Anilm3:
This PR adds a new preprocessor which parses a URI and extracts its constituent components into a new address. This new address can then be used to evaluate specific parts of a URL without over-complicated and inaccurate regular expressions. An example definition of this preprocessor in configuration can be seen below:

```json
{
  "id": "processor-001",
  "generator": "uri_parse",
  "conditions": [],
  "parameters": {
    "mappings": [
      {
        "inputs": [
          {
            "address": "server.io.net.url"
          }
        ],
        "output": "server.io.net.request.url"
      }
    ]
  },
  "evaluate": true,
  "output": false
}
```

The output of this preprocessor is a map with the following contents:

```yaml
{
  "scheme": <string>,
  "userinfo": <string>,
  "host": <string>,
  "port": <unsigned>,
  "path": <string>,
  "query": {},
  "fragment": <string>
}
```

Related Jiras: [APPSEC-58778]

[APPSEC-58778]: https://datadoghq.atlassian.net/browse/APPSEC-58778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/processor/uri_parse.cpp
+++ b/src/processor/uri_parse.cpp
@@ -14,7 +14,6 @@
 #include "uri_utils.hpp"
 #include "utils.hpp"
 
-#include <cstdint>
 #include <string_view>
 #include <unordered_map>
 #include <utility>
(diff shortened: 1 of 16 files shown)


src/processor/uri_parse.cpp
```diff
@@ -0,0 +1,162 @@
...
+    if (!decomposed.has_value()) {
+        return {};
+    }
+
+    ddwaf_object scheme = string_view_to_object(decomposed->scheme);
+    ddwaf_object userinfo = string_view_to_object(decomposed->authority.userinfo);
+    ddwaf_object host = string_view_to_object(decomposed->authority.host);
+
+    ddwaf_object port;
+    if (auto [res, value] = from_string<uint16_t>(decomposed->authority.port); res) {
```
cataphract, line 139: you're calling `from_string` already when parsing, though you discard the value; in any case, this can never fail if the string is non-empty. maybe you could add the port to decomposed in number form already.

cataphract (approved):

PR #441, opened 2025-08-18
Support other scalar types on action parameters

Anilm3:
This PR extends the set of allowed types on action parameters to include all other scalars. This is a small improvement which can be taken advantage of to prevent intermediate type conversions. In v2, other complex types will also be supported.

Related Jiras: [APPSEC-58779]

[APPSEC-58779]: https://datadoghq.atlassian.net/browse/APPSEC-58779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schema/actions.json
+++ b/schema/actions.json
@@ -9,13 +9,13 @@
       "type": "object",
       "properties": {
         "status_code": {
-          "type": "string"
+          "type": "number"
         },
         "type": {
           "type": "string"
         },
         "grpc_status_code": {
-          "type": "string"
+          "type": "number"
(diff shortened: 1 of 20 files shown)


cataphract (approved):
Looks good enough, but I wonder what the point is of having an additional tagged union representation that you create from ddwaf_object (and then back), when ddwaf_object is already a tagged union capable of representing your values.

PR #442, opened 2025-08-19
Subcontext: refactor evaluation stages from context

Anilm3:
Subcontexts will be implemented by:
- Adding and enforcing a scope (`context`, `subcontext`) to the evaluation process
- Adding a scope to the data (replacing `attribute::none` and `attribute::ephemeral`)
- Creating a new `subcontext` and `subcontext_wrapper` class which will be created from the current context (and with a linked lifecycle)

To do this, the evaluation process must be refactored out of the context, so that it can be used separately within the subcontext using the correct scope. This PR does exactly this first step, splitting these from `context` into the `evaluation_engine`.

The next PR will:
- Implement the subcontext piggybacking on ephemerals.
- Remove the concept of ephemerality in favour of scope (`context`, `subcontext`).
- Split the API to create and use subcontexts.

### DIFF
--- a/src/evaluation_engine.cpp
+++ b/src/evaluation_engine.cpp
@@ -2,17 +2,16 @@
 // dual-licensed under the Apache-2.0 License or BSD-3-Clause License.
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
-// Copyright 2021 Datadog, Inc.
-#include <chrono>
+// Copyright 2025 Datadog, Inc.
+
 #include <cstddef>
-#include <cstdint>
 #include <memory>
 #include <string_view>
 #include <utility>
 #include <vector>
 
 #include "clock.hpp"
-#include "context.hpp"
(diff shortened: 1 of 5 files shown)


src/evaluation_engine.cpp
```diff
@@ -50,65 +49,63 @@ void set_context_event_address(object_store &store)
 
 } // namespace
 
-std::pair<bool, owned_object> context::eval(uint64_t timeout)
+std::pair<bool, owned_object> evaluation_engine::eval(object_store &store, timer &deadline)
 {
     // This scope ensures that all ephemeral and cached objects are removed
     // from the store at the end of the evaluation
-    auto store_cleanup_scope = store_.get_eval_scope();
+    auto storecleanup_scope = store.get_eval_scope();
```
Anilm3, line 56: Will fix with any review comments or in a subsequent PR.

cataphract (approved):

PR #448, opened 2025-09-03
Support float status codes in configuration

Anilm3:
Following up on https://github.com/DataDog/libddwaf/pull/441, this PR adds support for validating status codes of type `float` (actually `double`).

### DIFF
--- a/src/configuration/actions_parser.cpp
+++ b/src/configuration/actions_parser.cpp
@@ -36,24 +36,41 @@ bool validate_status_code_presence_and_type(
         return false;
     }
 
-    if (holds_alternative<int64_t>(it->second)) {
-        if (const int64_t value = std::get<int64_t>(it->second); value >= 0) {
-            it->second = static_cast<uint64_t>(value);
+    auto validate_status_code = [&validate_fn]<typename T>(T code) {
+        auto ucode = static_cast<uint64_t>(code);
+        if constexpr (std::is_same_v<T, double>) {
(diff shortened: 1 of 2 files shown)


src/configuration/actions_parser.cpp
```diff
@@ -36,24 +36,41 @@ bool validate_status_code_presence_and_type(
...
+        code = static_cast<uint64_t>(std::get<double>(code));
+        return true;
+    }
+
+    if (holds_alternative<std::string>(code)) {
+        if (auto [res, value] = from_string<uint64_t>(std::get<std::string>(code));
+            res && validate_status_code(value)) {
+            code = value;
+            return true;
+        }
```
cataphract, line 69: you could maybe simplify things a bit:

```cpp
auto validate_status_code = [&validate_fn](auto&& code) {
    using T = std::decay_t<decltype(code)>;
    int64_t ucode;
    if constexpr (std::is_same_v<T, std::string>) {
        if (auto [res, value] = from_string<uint64_t>(code); res) {
            ucode = value;
        } else {
            return false;
        }
    } else {
        ucode = static_cast<uint64_t>(code);
        if (code < 0 || static_cast<T>(ucode) != code) {
            return false;
        }
    }
    return ucode <= 999 && validate_fn(ucode);
};

auto &code = it->second;
if (std::visit(validate_status_code, code)) {
    return true;
}
```

cataphract (approved):

PR #455, opened 2025-09-18
[v2] Return DDWAF_MATCH when there are events, attributes or actions

Anilm3:
This PR expands the cases in which `DDWAF_MATCH` is returned to include those in which attributes and actions are provided.


cataphract (approved):
This is a breaking change to the public API, so it should go on UPGRADING

PR #461, opened 2025-09-27
Incremental processor overrides

Anilm3:
This PR updates processor overrides to the new definition which adds or removes scanners to a given processor rather than fully replacing the whole set. The existing implementation is not currently in use, therefore there is no risk for existing users. More information can be found in RFC-1071.

An example processor override, adding and removing scanners, can be seen below:

```json
{
  "processor_overrides": [
    {
      "target": [{ "id": "extract-headers" }, { "id": "extract-content" }],
      "scanners": {
        "include": [
          {
            "tags": {
              "category": "pii"
            }
          },
          { "id": "mastercard-2x2" }
        ],
        "exclude": [{ "id": "mastercard-4x4" }]
      },
    }
  ]
}

```

Related Jiras: [APPSEC-59429]

[APPSEC-59429]: https://datadoghq.atlassian.net/browse/APPSEC-59429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/builder/processor_builder.cpp
+++ b/src/builder/processor_builder.cpp
@@ -8,7 +8,6 @@
 #include <set>
 #include <string>
 #include <utility>
-#include <vector>
 
 #include "builder/processor_builder.hpp"
 #include "configuration/common/configuration.hpp"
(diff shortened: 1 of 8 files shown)


Anilm3:
@codex review

src/builder/processor_builder.cpp
```diff
@@ -23,20 +22,50 @@
...
 std::set<const scanner *> references_to_scanners(
-    const std::vector<reference_spec> &references, const indexer<const scanner> &scanners)
+    const auto &references, const indexer<const scanner> &scanners)
 {
     std::set<const scanner *> scanner_refs;
-    for (const auto &ref : references) {
+
+    // Precedence:
+    //  - Refs by ID have precedence over refs by tags
+    //  - Exclusions have precedence over inclusions
```
cataphract, line 33: Could be worded more unambiguously. Does an inclusion by id have precedence over an exclusion by tags?

(From the code I infer that the answer is yes)

cataphract (approved):

PR #466, opened 2025-10-03
Pass correct capacity when moving cow_string to dynamic_string

Anilm3:


Anilm3:
@codex review

src/cow_string.hpp
```diff
@@ -103,20 +105,21 @@ class cow_string {
 
     // Moves the contents and invalidates the string if the buffer has been
     // modified, otherwise it does nothing
```
cataphract, line 107: The comment does not apply. It always invalidates the string. In fact, the conversion operator should be called on rvalue references only, so that the compiler can warn on uses after move.

cataphract (approved):

PR #468, opened 2025-10-10
Support multiple subcontexts per context

Anilm3:
This PR changes the way subcontexts are used and created, they are now a "copy" of the context, allowing the creation and use of many subcontexts at the same time, although not in a thread-safe manner without user-provided synchronisation. This iteration has some caveats:
- The context still owns the data provided to it, so the subcontext just copies the targets but points to the context data.
- The cache is fully copied, but in this iteration:
  - Matches performed with context data are not written back to the context cache.
  - Unevaluated primitives may be evaluated within the subcontext even if their outcome relates to the context.

All of the logic to distinguish contexts and subcontexts internally has been removed for simplification, albeit some of it may come back in the future to address the caveats above.

Remaining work (to be addressed in a future PR):
* [x] Evaluation within the context must be forbidden if a subcontext exists to avoid inconsistencies, this hasn't been addressed here.
* [x] If a context is destroyed while there is a subcontext alive, the user-provided data is invalidated, this hasn't been addressed either.
* [x] There are many unit tests for contexts and subcontexts which are no longer relevant, these have to be removed.
* [ ] There are no new tests to validate the "concurrency-without-thread-safety" aspect as the PR is already too large.

Related Jiras: [APPSEC-59671]

[APPSEC-59671]: https://datadoghq.atlassian.net/browse/APPSEC-59671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Anilm3:
@codex review

cataphract (approved):

PR #486, opened 2025-12-24
Remove default allocator arguments

cataphract:
The default allocator argument caused runtime exceptions "emplace: incompatible allocators" (from writable_object::emplace/emplace_back in src/object.hpp:1179, 1216) when converting a dynamic_string to an object with a mismatched allocator.

By making the allocator parameter required, such bugs will be caught at compile time instead of causing runtime exceptions.

This change fixes two instances in fingerprint.cpp where the allocator was not being passed, which triggered these allocator mismatch exceptions, and updates tests to explicitly pass allocators.

### DIFF
--- a/src/object.hpp
+++ b/src/object.hpp
@@ -824,8 +824,8 @@ protected:
 class owned_object final : public readable_object<owned_object>,
                            public writable_object<owned_object> {
 public:
-    explicit owned_object(nonnull_ptr<memory::memory_resource> alloc)
-        : obj_({.type = object_type::invalid}), alloc_(alloc)
+    owned_object()
+        : obj_({.type = object_type::invalid}), alloc_(memory::get_default_null_resource())
     {}
 
     // UNSAFE: Caller must ensure the object's memory is compatible with the
@@ -863,11 +863,6 @@ public:
     [[nodiscard]] detail::object *ptr() { return &obj_; }
     [[nodiscard]] const detail::object *ptr() const noexcept { return &obj_; }
     [[nodiscard]] nonnull_ptr<memory::memory_resource> alloc() const noexcept { return alloc_; }
-
-    static owned_object make_invalid()
-    {
-        return owned_object({.type = object_type::invalid}, memory::get_default_null_resource());
-    }
 
     // the variants that don't take a memory resource can't be written to
     // (have their value replaced)
@@ -1085,7 +1080,7 @@ template <typename Derived>
         default:
             break;
         }
-        return owned_object{alloc};
+        return owned_object{};
     };
 
     std::deque<std::pair<object_view, borrowed_object>> queue;
--- a/src/object.hpp
+++ b/src/object.hpp
@@ -391,8 +391,7 @@ public:
     // Convert the underlying type to the requested type
     template <typename T> T convert() const;
 
-    [[nodiscard]] owned_object clone(
-        nonnull_ptr<memory::memory_resource> alloc = memory::get_default_resource()) const;
+    [[nodiscard]] owned_object clone(nonnull_ptr<memory::memory_resource> alloc) const;
 
 private:
     readable_object() = default;
@@ -790,20 +789,6 @@ private:
 class borrowed_object final : public readable_object<borrowed_object>,
                               public writable_object<borrowed_object> {
 public:
-    explicit borrowed_object(detail::object *obj,
-        nonnull_ptr<memory::memory_resource> alloc = memory::get_default_resource())
-        : obj_(obj), alloc_(alloc)
-    {
-        if (obj_ == nullptr) {
-            throw std::invalid_argument("invalid borrowed object (null)");
-        }
-    }
-
-    explicit borrowed_object(detail::object &obj,
-        nonnull_ptr<memory::memory_resource> alloc = memory::get_default_resource())
-        : obj_(&obj), alloc_(alloc)
-    {}
-
     explicit borrowed_object(owned_object &obj);
     borrowed_object &operator=(owned_object &&obj);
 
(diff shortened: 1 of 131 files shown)


cataphract:
There were more similar bugs. Added another commit to remove more default allocator parameters

src/object.hpp
```diff
@@ -813,75 +798,48 @@ class borrowed_object final : public readable_object<borrowed_object>,
...
+        }
+        return borrowed_object{obj, alloc};
+    }
+
+    // UNSAFE: Caller must ensure the object's memory is compatible with the allocator
+    static borrowed_object create_unchecked(
+        detail::object &obj, nonnull_ptr<memory::memory_resource> alloc)
+    {
+        return borrowed_object{&obj, alloc};
+    }
```
Anilm3: These provide no value when compared to the removed constructors, `checked` / `unchecked`  function semantics / naming is not really something I want to introduce. Any specifics should be left to documentation, so unless there's something I'm missing, you should add the constructors back, remove the default allocator argument and keep the comments.

src/object.hpp
```diff
@@ -900,61 +858,91 @@ class owned_object final : public readable_object<owned_object>,
...
+            {.type = object_type::invalid}, memory::get_default_null_resource());
+    }
 
-    static owned_object make_boolean(bool value)
+    // variant of make_invalid() for subsequent assignment
+    static owned_object make_uninit(nonnull_ptr<memory::memory_resource> alloc)
     {
-        return owned_object{{.via{.b8{.type = object_type::boolean, .val = value}}}};
+        return create_unchecked({.type = object_type::invalid}, alloc);
     }
```
Anilm3: As mentioned, these are not needed, just reintroduce the default constructor.

src/object.hpp
```diff
@@ -900,61 +858,91 @@ class owned_object final : public readable_object<owned_object>,
...
+            alloc);
+    }
+
+    // UNSAFE: Does not copy the string - caller must ensure the string memory
+    // can be deallocated by alloc
+    static owned_object unsafe_make_string_nocopy(
+        std::same_as<std::string_view> auto str, nonnull_ptr<memory::memory_resource> alloc)
+    {
+        return unsafe_make_string_nocopy(str.data(), str.size(), alloc);
+    }
```
Anilm3: I don't want to start labelling functions based on safety / unsafety as it's a slippery slope, the comments are good enough for that purpose. Please remove the `unsafe_` prefix from these functions.

src/object.hpp
```diff
@@ -963,69 +951,84 @@ class owned_object final : public readable_object<owned_object>,
...
         return copy;
     }
 
+    // UNSAFE: Caller must ensure the object's memory is compatible with the
+    // allocator. In particular:
+    // 1) if it needs to be deallocated, alloc must be able to do so
+    // 2) if it is a map or array that requires resizing, allow must be able to
+    //    free the current chunk and allocate a new one
+    // 3) if it contains sub-objects, the requirements apply transitively
+    static owned_object create_unchecked(
```
Anilm3: This function is no more useful than the constructor that was removed, in addition the checked / unchecked suffix is not something I want to introduce. Please reintroduce the public explicit constructor, but keep the comments.

src/object.hpp
```diff
@@ -900,61 +858,91 @@ class owned_object final : public readable_object<owned_object>,
...
+    static owned_object make_float(double value,
+        nonnull_ptr<memory::memory_resource> alloc = memory::get_default_null_resource())
     {
-        return owned_object{{.via{.str{.type = object_type::string,
-                                .size = len,
-                                // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
-                                .ptr = const_cast<char *>(str)}}},
-            alloc};
+        return create_unchecked({.via{.f64{.type = object_type::float64, .val = value}}}, alloc);
     }
```
Anilm3: As discussed allocators are not needed here as an argument, just pass the null allocator to the constructor (which will replace `create_unchecked`. The assignment should work as expected in those cases as well.

src/object.hpp
```diff
@@ -963,69 +951,84 @@ class owned_object final : public readable_object<owned_object>,
...
 
-        return owned_object{
+        return create_unchecked(
             {.via{.map{.type = object_type::map,
                 .size = 0,
                 .capacity = capacity,
                 .ptr = detail::alloc_helper<detail::object_kv, uint16_t>(capacity, *alloc)}}},
-            alloc};
+            alloc);
     }
```
Anilm3, line 1002: I would also add `make_map(allocator)` and `make_array(allocator)` to avoid having to specify `0` in those cases.

src/object.hpp
cataphract: this takes an allocator so a default constructor would not be appropriate. In fact, there is little use for a default constructor -- it could be used only for subsequent assignment of a scalar, which you might as well do with `create_x`

src/object.hpp
cataphract: The allocators _may_ be needed, in case these are reassigned. Because in most circumstances they aren't, a _safe_ (i.e. if you screw up it doesn't crash, just throws during execution) default value is provided.

Anilm3: The allocator is not needed. The assignment will attempt to free with the null allocator (noop) and replace the allocator as needed.

src/object.hpp
Anilm3: This takes an allocator that isn't needed.

src/object.hpp
cataphract: hmm I hadn't noticed owned_object had its own assignment operator that replaced the allocator. I'll make the changes you request, but I'll note that implicitly changing the allocator is questionable (e.g it implicitly changes the possible valid lifetime bounds of the variable to that of the new allocator), and in fact PMR, which we're effectively emulating here, doesn't do it.

src/object.hpp
Anilm3: Not sure if it was clear, but this is no longer needed, the default constructor must result in an invalid object.

src/object.hpp
```diff
@@ -825,54 +824,19 @@ class borrowed_object final : public readable_object<borrowed_object>,
 class owned_object final : public readable_object<owned_object>,
                            public writable_object<owned_object> {
 public:
-    owned_object() = default;
-    explicit owned_object(detail::object obj,
-        nonnull_ptr<memory::memory_resource> alloc = memory::get_default_resource())
-        : obj_(obj), alloc_(alloc)
+    explicit owned_object(nonnull_ptr<memory::memory_resource> alloc)
+        : obj_({.type = object_type::invalid}), alloc_(alloc)
     {}
```
Anilm3: I think this constructor is also not needed, you can fold both make_invalid and this constructor into the default constructor which generates an invalid object with the null constructor.

Anilm3 (approved):

PR #487, opened 2026-01-20
Release v2.0-alpha0 and documentation

Anilm3:
This PR includes the following changes:
- The coverage upload jobs have been split to avoid the deployment errors on non-master branches.
- The release job should now automatically upload release notes.
- Relevant documentation has now been moved under docs/
- Some documentation has been written through codex and claude, albeit most of it is preliminary and needs extensive work before considered accurate enough (this will be done over time...).
- The following docs are stable:
  - `docs/changelog/*`
  - `docs/upgrading/*`
  - `docs/allocators.md`


docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Library Lifecycle Functions
+
+### `ddwaf_init`
+
+```c
+ddwaf_handle ddwaf_init(const ddwaf_object *ruleset, ddwaf_object *diagnostics);
+```
+
+Creates a WAF instance from a configuration object. The `ruleset` must be a map that contains the rule definitions, overrides, data, and exclusions. When supplied, `diagnostics` receives parser warnings and errors; destroy it with `ddwaf_object_destroy` after use. Returns `NULL` on failure.
```
cataphract: Doesn't indicate which allocator to use with `ddwaf_object_destroy` (the default allocator).

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+
+Returns a null-terminated string describing the libddwaf version. The pointer remains owned by the library.
+
+### `ddwaf_set_log_cb`
+
+```c
+bool ddwaf_set_log_cb(ddwaf_log_cb cb, DDWAF_LOG_LEVEL min_level);
+```
+
+Installs a callback that receives libddwaf log messages at or above `min_level`. Passing `NULL` disables logging. Returns `false` if the logging system rejects the callback (for example, when called concurrently).
```
cataphract: It always returns true, actually. It should note that it writes to a global without synchronization so in effect it means the caller must externally synchronize (e.g. by setting it before starting threads that will call the WAF, by using a bool guard with release-acquire, etc.)

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+
+Evaluates a request payload. `data` must be a map keyed by address. The `alloc` parameter is the allocator used to free the `data` object after evaluation completes (if non-null). When `result` is non-null and the return code is `DDWAF_OK` or `DDWAF_MATCH`, it contains detailed events and metadata allocated with the `output_alloc` from `ddwaf_context_init`, and must be released with `ddwaf_object_destroy(result, output_alloc)`. The function enforces a microsecond timeout.
+
+### `ddwaf_context_destroy`
+
+```c
+void ddwaf_context_destroy(ddwaf_context context);
+```
+
+Destroys a context and releases any persistent copies of request data using the allocator associated with the context.
```
cataphract: I don't think this is correct, It releases the copies of the request data using the allocator associated with each bout of data passed via context_eval -- not with the allocator "associated with the context" (the one passed to `ddwaf_context_init`).

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+```c
+DDWAF_RET_CODE ddwaf_subcontext_eval(
+    ddwaf_subcontext subcontext,
+    ddwaf_object *data,
+    ddwaf_allocator alloc,
+    ddwaf_object *result,
+    uint64_t timeout);
+```
+
+Evaluates request data against a subcontext using the same semantics as `ddwaf_context_eval`. The `alloc` parameter is used to free the `data` object. The `result` object, when populated, is allocated with the `output_alloc` from the parent context's `ddwaf_context_init` call and must be destroyed with that same allocator.
```
cataphract: could indicate that alloc is nullable

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+```c
+bool ddwaf_builder_add_or_update_config(
+    ddwaf_builder builder,
+    const char *path,
+    uint32_t path_len,
+    const ddwaf_object *config,
+    ddwaf_object *diagnostics);
+```
+
+Registers or refreshes a configuration identified by `path`. The `config` object is parsed immediately; on error the return value is `false` and `diagnostics` (when provided) contains parser feedback.
```
cataphract: it should indicate how `diagnostics` should be freed: `ddwaf_object_destroy(..., ddwaf_get_default_allocator())`.

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+
+```c
+uint32_t ddwaf_builder_get_config_paths(
+    ddwaf_builder builder,
+    ddwaf_object *paths,
+    const char *filter,
+    uint32_t filter_len);
+```
+
+Counts the configurations present in the builder and optionally populates `paths` with their identifiers. If `filter` is non-null, only the paths matching the unanchored regular expression are reported.
```
cataphract: no indication of the lifetime of `paths`

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Allocator APIs
+
+### `ddwaf_get_default_allocator`
+
+```c
+ddwaf_allocator ddwaf_get_default_allocator(void);
+```
+
+Returns the allocator used internally by libddwaf. The handle is owned by the library and must not be destroyed.
```
cataphract: This seems to contradict a later comment that says it's a noop.

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Allocation helpers
+
+```c
+void *ddwaf_allocator_alloc(ddwaf_allocator alloc, size_t bytes, size_t alignment);
+void ddwaf_allocator_free(ddwaf_allocator alloc, void *p, size_t bytes, size_t alignment);
+void ddwaf_allocator_destroy(ddwaf_allocator alloc);
+```
+
+Low-level allocation routines. The `alloc` argument must be the same handle returned by the allocator factory. `ddwaf_allocator_destroy` is a no-op for the default allocator but must be called for user-created allocators once all objects have been released.
```
cataphract: > user-created allocators

not clear if this covers `ddwaf_synchronized_pool_allocator_init` et al

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+### Container insertion and JSON parsing
+
+```c
+ddwaf_object *ddwaf_object_insert(ddwaf_object *array, ddwaf_allocator alloc);
+ddwaf_object *ddwaf_object_insert_key(ddwaf_object *map, const char *key, uint32_t length, ddwaf_allocator alloc);
+ddwaf_object *ddwaf_object_insert_literal_key(ddwaf_object *map, const char *key, uint32_t length, ddwaf_allocator alloc);
+ddwaf_object *ddwaf_object_insert_key_nocopy(ddwaf_object *map, const char *key, uint32_t length, ddwaf_allocator alloc);
+bool ddwaf_object_from_json(ddwaf_object *output, const char *json_str, uint32_t length, ddwaf_allocator alloc);
+```
+
```
cataphract, line 564: it should be indicated that all the same allocator should be used throughout an object graph. Arguably, instead of taking a returning a ddwaf_object*, these should take and return a fat pointer with the allocator

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+### Container lookup
+
+```c
+const ddwaf_object *ddwaf_object_at_key(const ddwaf_object *object, size_t index);
+const ddwaf_object *ddwaf_object_at_value(const ddwaf_object *object, size_t index);
+const ddwaf_object *ddwaf_object_find(const ddwaf_object *object, const char *key, size_t length);
+```
+
+`ddwaf_object_at_key` returns the key object for the entry at `index` when iterating over a map (arrays do not expose keys). `ddwaf_object_at_value` returns the corresponding value object. `ddwaf_object_find` performs a direct lookup by key and returns the associated value.
```
cataphract: does not indicate whether `ddwaf_object_at_key/value` has bounds checking or whether `ddwaf_object_find` returns NULL if not found.

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+void ddwaf_object_destroy(ddwaf_object *object, ddwaf_allocator alloc);
+```
+
+Reclaims any memory owned by the object. Always pass the allocator that was used to build it.
+
+---
+
+## Thread Safety and Lifetime Notes
+
+- WAF handles are thread-safe for read operations (`ddwaf_context_init`, `ddwaf_known_addresses`, `ddwaf_known_actions`). Builder operations and handle destruction are not thread-safe with respect to these read operations.
```
cataphract: > are not thread-safe with respect to these read operations

It's not clear to me what this means.

docs/c-api/api.md
```diff
@@ -0,0 +1,441 @@
...
+- Allocation routines inherit the thread-safety guarantees of the allocator instance returned by the corresponding factory.
+- Objects created with `_string_literal` or `_insert_literal_key` expect the referenced buffers to outlive the object or to remain immutable.
```
cataphract: In fact, they expect them to remain immutable because they also store their size. Also, if they're not immutable, there will be thread safety concerns.

docs/upgrading/UPGRADING-v2.0.md
```diff
@@ -0,0 +1,537 @@
...
+
+**v1.x:**
+```c
+ddwaf_context context = ddwaf_context_init(handle);
+
+// Store persistent data
+ddwaf_object persistent = ...;
+ddwaf_result result1;
+ddwaf_run(context, &persistent, NULL, &result1, timeout);
+ddwaf_result_free(&result1);
```
cataphract: ddwaf_result_free is no more

cataphract (commented):
unrelated, but infer_schema.cpp also needs to be looked at

docs/upgrading/UPGRADING-v2.0.md
Anilm3: Good catch

cataphract (approved):

PR #493, opened 2026-05-18
Add `ddwaf_object_stringl` and `ddwaf_object_map_addl`

KowalskiThomas:
## What is this PR?

Needed for https://github.com/DataDog/dd-trace-py/pull/18128.

### DIFF
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -674,6 +674,17 @@ ddwaf_object* ddwaf_object_set_null(ddwaf_object *object);
 ddwaf_object* ddwaf_object_set_string(ddwaf_object *object, const char *string, uint32_t length, ddwaf_allocator alloc);
 
 /**
+ * Backward-compatible wrapper: creates a string object using the default allocator.
+ *
+ * @param object Object to perform the operation on. (nonnull)
+ * @param string String to initialise the object with, this string will be copied. (nonnull)
+ * @param length Length of the string.
+ *
+ * @return A pointer to the passed object or NULL if the operation failed.
+ **/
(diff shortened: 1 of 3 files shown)


Anilm3 (commented):
These helpers are from libddwaf v1 and so is `ddwaf_object_map_addl` and have been deprecated, `master` is in v2.

KowalskiThomas:
@Anilm3 the reason why I initially made this PR is that I got a security report saying some payloads might be ignored by the WAF due to `strlen` stopping at the first null byte, which might not be the actual end of the string as far as Python is concerned.

```py
>>> s = "0\x00thomas"
>>> print(s)
0thomas
>>> print(len(s))
8
```

Do you think the library/consumer-side should handle those cases before calling into the WAF? What would be the recommended workaround?

Anilm3:
> @Anilm3 the reason why I initially made this PR is that I got a security report saying some payloads might be ignored by the WAF due to `strlen` stopping at the first null byte, which might not be the actual end of the string as far as Python is concerned.
> 
> ```python
> >>> s = "0\x00thomas"
> >>> print(s)
> 0thomas
> >>> print(len(s))
> 8
> ```
> 
> Do you think the library/consumer-side should handle those cases before calling into the WAF? What would be the recommended workaround?

@KowalskiThomas that's not really the issue I was trying to raise, the object constructors let you follow either approach. The issue with this PR is that your python code is likely using libddwaf v1, while master is on v2, therefore those helpers are not necessary. In addition, the helpers already exist in libddwaf v1.

KowalskiThomas:
Thanks for the comment and explanation -- I discussed with the Python folks and this isn't a problem for Python anyway, so I'll just close everything. Very sorry for the noise.

PR #494, opened 2026-06-09
Introduce ddwaf_(context|subcontext)_multieval to evaluate multiple batches in sequence

Anilm3:
## Overview

This PR adds two new public entrypoints, `ddwaf_context_multieval` and `ddwaf_subcontext_multieval`, that evaluate multiple input batches in sequence within a single call and return one combined result. Functionally each entrypoint behaves like its `_eval` counterpart, except that `data` is an array of maps where every element is treated as a separate input batch and evaluated in order.

To support this, insertions to the object store now result in the batch being added to a queue, which then has to be explicitly applied before it's ready to be evaluated:

- `insert_batch` / `insert_batches` enqueue input (a single map, or an array of maps) without applying it.
- `next_batch` pops and applies the next queued batch, marking its targets as new.
- `flush_input_queue` drains any batches left unevaluated (e.g. after a timeout) on every exit path, applying them as "existing" targets targets so they carry over to subsequent calls and resets the new-target set for the next evaluation.

The evaluation engine now loops over batches, running pre-processors, filters, rules, and post-processors per batch, accumulating events/actions/attributes into the single result.

A new field has been added to the result object:

- **`evaluated`**: an unsigned integer reporting how many batches were fully evaluated. In the normal case this equals the number of non-empty batches; on timeout/error during batch `i` (0-based, counting non-empty batches) it equals `I`, i.e. the index of the batch where the problem occurred. Empty batches are skipped and don't count.

This field is present on the regular `_eval` results too (the result schema is shared), where it's simply `1` for a non-empty evaluated batch or `0` otherwise.

## Files to review (`src/`)

- **`src/object_store.hpp`** / **`src/object_store.cpp`**: core change: the batch queue, `insert_batch`/`insert_batches`/`insert_target`/`next_batch`/`flush_input_queue`/`enqueue_batch`/`apply_batch`, and the `insert_and_apply` test helper. Worth the closest look for the queue lifetime and new-target semantics.
- **`src/evaluation_engine.cpp`**: the per-batch evaluation loop, `flush_input_queue` on exit, and the `evaluated` counter.
- **`src/evaluation_engine.hpp`**: `insert_batch`/`insert_batches` forwarding and the `next_batch`/`insert_and_apply` test internals.
- **`src/interface.cpp`**: the two new public functions plus the array-vs-map dispatch added to `ddwaf_context_eval` / `ddwaf_subcontext_eval`.
- **`src/context.hpp`**: `context`/`subcontext` forwarding for the new store methods and test internals.
- **`src/serializer.hpp`** / **`src/serializer.cpp`**: the `evaluated` field added to `result_components` and `initialise_result_object`.
- **`src/processor/base.hpp`**: processor output switched from `insert` to `insert_target`.

### DIFF
--- a/src/evaluation_engine.cpp
+++ b/src/evaluation_engine.cpp
@@ -5,11 +5,16 @@
 // Copyright 2025 Datadog, Inc.
 
 #include <cstddef>
+#include <cstdint>
+#include <iterator>
 #include <memory>
+#include <string>
 #include <string_view>
 #include <utility>
+#include <variant>
 #include <vector>
 
+#include "attribute_collector.hpp"
 #include "clock.hpp"
 #include "evaluation_engine.hpp"
 #include "exception.hpp"
@@ -42,7 +47,34 @@ void set_context_event_address(object_store &store)
         return;
     }
 
-    store.insert_target(event_addr_idx, event_addr, owned_object::make_boolean(true));
+    store.insert_and_apply(event_addr_idx, event_addr, owned_object::make_boolean(true));
+}
+
+void collect_attributes(const object_store &store, const std::vector<rule_result> &results,
+    attribute_collector &collector)
+{
+
+    // First collect any pending attributes from previous runs
+    collector.collect_pending(store);
+
+    for (const auto &result : results) {
+        for (const auto &attr : result.attributes.get()) {
+            if (std::holds_alternative<rule_attribute::input_target>(attr.value_or_target)) {
+                auto input = std::get<rule_attribute::input_target>(attr.value_or_target);
+                collector.collect(store, input.index, input.key_path, attr.key);
+            } else if (std::holds_alternative<std::string>(attr.value_or_target)) {
+                collector.insert(attr.key, std::get<std::string>(attr.value_or_target));
+            } else if (std::holds_alternative<uint64_t>(attr.value_or_target)) {
+                collector.insert(attr.key, std::get<uint64_t>(attr.value_or_target));
+            } else if (std::holds_alternative<int64_t>(attr.value_or_target)) {
+                collector.insert(attr.key, std::get<int64_t>(attr.value_or_target));
+            } else if (std::holds_alternative<double>(attr.value_or_target)) {
+                collector.insert(attr.key, std::get<double>(attr.value_or_target));
+            } else if (std::holds_alternative<bool>(attr.value_or_target)) {
+                collector.insert(attr.key, std::get<bool>(attr.value_or_target));
+            }
+        }
+    }
 }
 
 } // namespace
@@ -57,14 +89,17 @@ std::pair<bool, owned_object> evaluation_engine::eval(timer &deadline)
     // Once evaluation finishes (on any exit path, including a timeout) flush any
     // input batches left unevaluated and reset the new-target set so that the
     // next eval can identify new targets.
-    auto on_exit = defer([this]() { store_.flush_input_queue(); });
+    auto on_exit = defer([this]() { input_batches_.flush(store_); });
 
-    std::vector<rule_result> results;
+    std::vector<rule_result> all_results;
     std::size_t batches_evaluated = 0;
+    bool event_addr_added = false;
 
     // Each queued input batch is evaluated as if it were a separate eval call,
-    // draining the store's queue one batch at a time.
-    while (store_.next_batch()) {
+    // draining the queue one batch at a time.
+    while (input_batches_.next_batch(store_)) {
+        std::vector<rule_result> results;
+        auto verdict = rule_verdict::none;
         try {
             // Evaluate preprocessors first in their own try-catch, if there's a
             // timeout we still need to evaluate rules unaffected by it.
@@ -84,16 +119,27 @@ std::pair<bool, owned_object> evaluation_engine::eval(timer &deadline)
                 const auto &policy = eval_filters(deadline);
 
                 if (should_eval_rules) {
-                    eval_rules(policy, results, deadline);
-                    if (!results.empty()) {
+                    verdict = eval_rules(policy, results, deadline);
+                    if (!event_addr_added && !results.empty()) {
                         set_context_event_address(store_);
+                        event_addr_added = true;
                     }
                 }
             }
 
             eval_postprocessors(deadline);
             ++batches_evaluated;
-        } catch (const ddwaf::timeout_exception &) {
+            // NOLINTNEXTLINE(bugprone-empty-catch)
+        } catch (const ddwaf::timeout_exception &) {}
+
+        collect_attributes(store_, results, collector_);
+
+        if (!results.empty()) {
+            all_results.insert(all_results.end(), std::make_move_iterator(results.begin()),
+                std::make_move_iterator(results.end()));
+        }
+
+        if (deadline.expired() || verdict == rule_verdict::block) {
             break;
         }
     }
@@ -104,7 +150,7 @@ std::pair<bool, owned_object> evaluation_engine::eval(timer &deadline)
     // available (e.g. from a postprocessor) and return a map of all attributes
     // generated during this call.
     // object::assign(result.attributes, collector_.collect_pending(store));
-    serializer.serialize(store_, results, collector_, deadline, output);
+    serializer.serialize(all_results, collector_, deadline, output);
     return {!output.attributes.empty() || !output.actions.empty() || !output.events.empty(),
         std::move(result_object)};
 }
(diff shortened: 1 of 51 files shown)


Anilm3:
@codex review

src/evaluation_engine.cpp
```diff
@@ -42,57 +42,63 @@ void set_context_event_address(object_store &store)
...
+
+                if (should_eval_rules) {
+                    eval_rules(policy, results, deadline);
+                    if (!results.empty()) {
+                        set_context_event_address(store_);
+                    }
                 }
             }
+
+            eval_postprocessors(deadline);
```
christophe-papazian, line 126: If eval_preprocessors times out (caught above) and the ruleset has no postprocessors, eval_postprocessors returns without ever checking the deadline — so batches_evaluated still increments. 

Shouldn't there be an early deadline check here, or at the top of eval_postprocessors, to guard against this?

tests/integration/interface/context/multieval/test.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    ddwaf_object data;
+    ddwaf_object_set_map(&data, 0, alloc);
+
+    ddwaf_object result;
+    ddwaf_object_set_invalid(&result);
+
+    EXPECT_EQ(ddwaf_context_multieval(context, &data, alloc, &result, LONG_TIME),
+        DDWAF_ERR_INVALID_OBJECT);
+
+    ddwaf_object_destroy(&data, alloc);
```
christophe-papazian: When alloc is non-null, insert_batches moves the object into input_objects_ before the is_array() check, so the context owns the memory even on the INVALID_OBJECT path. Calling ddwaf_object_destroy here would double-free it when the context is later destroyed, no?

tests/integration/interface/context/multieval/test.cpp
```diff
@@ -0,0 +1,1033 @@
...
+    ddwaf_object data;
+    ddwaf_object_set_map(&data, 0, alloc);
+
+    ddwaf_object result;
+    ddwaf_object_set_invalid(&result);
+
+    EXPECT_EQ(ddwaf_subcontext_multieval(subctx, &data, alloc, &result, LONG_TIME),
+        DDWAF_ERR_INVALID_OBJECT);
+
+    ddwaf_object_destroy(&data, alloc);
```
christophe-papazian: same here?

tests/unit/evaluation_engine_test.cpp
```diff
@@ -751,10 +751,10 @@ TEST(TestEvaluationEngine, RuleFilterWithSubcontextConditionMatch)
         auto persistent = object_builder_da::map({{"usr.id", "admin"}});
         auto ephemeral = object_builder_da::map({{"http.client_ip", "192.168.0.1"}});
 
-        EXPECT_TRUE(ctx.insert(std::move(persistent)));
+        EXPECT_TRUE(ctx.insert_batch(std::move(persistent)));
```
christophe-papazian: After this rename from insert() to insert_batch(), persistent is only queued, not applied to targets_. create_subcontext() snapshots targets_ (not the queue), so the subcontext eval runs without usr.id. Shouldn't this use insert_and_apply() to preserve the original behaviour of this test?

christophe-papazian (commented):
The serializer now always writes `evaluated` into the result object, but schema/result.json still has `additionalProperties: false` and doesn't list it. Worth updating the schema, or will this silently break consumers doing strict validation?

src/evaluation_engine.cpp
Anilm3: It's a bit of an edge-case that probably doesn't matter much as it'll timeout in a rule before reaching the filters. A few other things around the timeout have to be re-worked so I'll address that edge case later.

tests/integration/interface/context/multieval/test.cpp
Anilm3: This is true, but claude wrote this test without allocating any memory (and the one below).

tests/unit/evaluation_engine_test.cpp
Anilm3: Nice one.

estringana (approved):

src/evaluation_engine.cpp
```diff
@@ -42,57 +42,63 @@ void set_context_event_address(object_store &store)
...
+            const bool should_eval_filters = should_eval_rules || check_new_filter_targets();
+
+            if (should_eval_filters) {
+                // Filters need to be evaluated even if rules don't, otherwise it'll
+                // break the current condition cache mechanism which requires knowing
+                // if an address is new to this run.
+                const auto &policy = eval_filters(deadline);
+
+                if (should_eval_rules) {
+                    eval_rules(policy, results, deadline);
```
cataphract: the rfc says "The WAF will evaluate each map sequentially, with the same semantics as a sequence of calls with a single map. However the first _blocking rule triggered_, error or timeout will end the whole evaluation." (emphasis mine). I don't see it stopping here. To be clear, the evaluation of a module inside an eval_rules call will stop that eval_rules call, but the RFC says "whole evaluation", which I take to mean the evaluation of all the maps.

src/object_store.hpp
```diff
@@ -23,9 +23,48 @@ class object_store {
...
+        insert_batch(input);
+        next_batch();
+        return true;
+    }
+
+    // Apply any remaining queued batches as targets *without* marking them as
+    // new (so they won't be evaluated) and clear the new-target set. Invoked
+    // once evaluation finishes, on any exit path including a timeout, mirroring
+    // the inputs being carried over to a subsequent ddwaf_context_eval call.
+    void flush_input_queue();
```
cataphract: not sure bolting this on `object_store` is the correct choice. I'd create a new class

src/evaluation_engine.cpp
```diff
@@ -42,57 +42,63 @@ void set_context_event_address(object_store &store)
...
+                    if (!results.empty()) {
+                        set_context_event_address(store_);
+                    }
```
cataphract: results is kept across iterations, so this may be called repeatedly

src/evaluation_engine.cpp
cataphract: because the rule_result is not self-contained and the serializer is, perhaps surprisingly, responsible for non-serialization tasks like resolving attributes, I think this can cause attributes to be resolved against the wrong input, in case it's replaced later in the batch. This would not happen if you ran context_val several times in succession

src/object_store.cpp
```diff
@@ -14,26 +14,92 @@
...
+        return false;
+    }
+
+    return insert_batches(array_view{view});
+}
+
+void object_store::enqueue_batch(map_view input)
+{
+    // Batches with no addresses are considered valid as they are harmless
+    if (!input.empty()) {
```
cataphract: this likely has implications for `evaluated` in a way that's confusing for the caller

src/evaluation_engine.cpp
Anilm3: You're right, I missed that.

src/object_store.hpp
Anilm3: I was using an "iterator-like" class before but it wasn't much cleaner, I'll give it another thought.

src/evaluation_engine.cpp
Anilm3: I'll see if I can extract the attribute resolution from the serialisation. Otherwise, it's an edge-case we can probably live with...

It would be so much easier if the result was simply an array of results...

src/object_store.cpp
Anilm3: I mean... technically there's nothing to evaluate.

cataphract: yes, but the RFC says "Hence, in case of a timeout or an error, it will be the index of the step where the problem occurred."

Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/494/commits/9b1dbc72cf03275b4135ed88d0180e17d66aef59

src/evaluation_engine.cpp
Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/494/commits/9b1dbc72cf03275b4135ed88d0180e17d66aef59

src/evaluation_engine.cpp
Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/494/commits/9b1dbc72cf03275b4135ed88d0180e17d66aef59

src/object_store.hpp
Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/494/commits/9b1dbc72cf03275b4135ed88d0180e17d66aef59

src/evaluation_engine.cpp
Anilm3: Addressed in https://github.com/DataDog/libddwaf/pull/494/commits/9b1dbc72cf03275b4135ed88d0180e17d66aef59

Anilm3:
@codex review

cataphract (approved):

PR #500, opened 2026-06-28
logs: preserve relative path of source files

cataphract:
So we can properly link to them.


estringana:
I'm ok with the PR but what's the motivation behind?

cataphract:
@estringana 
> I'm ok with the PR but what's the motivation behind?

it's so that the file where the log call is made can be identified unambiguously. Discard the relative path to src removes needed information to e.g. automatically link to the relevant line.

estringana (approved):

PR #502, opened 2026-06-30
Use inline storage for value iterators

cataphract:
Dramatically reduce malloc contention under musl. glibc also has significant improvements. See commit message for more details.


Taiki-San (approved):
Looking good.

The pointer arithmetic around the access to inline_ptr is a little bit scary/brittle (I'm thinking about race conditions or new call sites without a boundary check) but the current implementation/usage  is fine

PR #506, opened 2026-07-30
Add support for transformers on RASP operators

Anilm3:
Overview
--
This PR introduces support for transformers on the request inputs provided to RASP operators (shi, cmdi, ssrf, sqli and lfi), for example, the following rule attempts to base64 decode all query parameters before performing a match against the provided resource:

```json
    {
      "id": "rasp-cmdi",
      "name": "Shell command injection exploit",
      "tags": {
        "type": "command_injection",
        "category": "vulnerability_trigger",
      },
      "conditions": [
        {
          "parameters": {
            "resource": [
              {
                "address": "server.sys.shell.cmd"
              }
            ],
            "params": [
              {
                "address": "server.request.query",
                "transformers": [ "base64_decode" ]
              }
            ]
          },
          "operator": "shi_detector"
        }
      ],
    },
```

Modifiers such as `keys_only` or `values_only` are not supported.

### DIFF
--- a/src/condition/cmdi_detector.cpp
+++ b/src/condition/cmdi_detector.cpp
@@ -31,7 +31,9 @@
 #include "object_type.hpp"
 #include "platform.hpp"
 #include "tokenizer/shell.hpp"
+#include "transformer/base.hpp"
 #include "transformer/lowercase.hpp"
+#include "transformer/manager.hpp"
 #include "utils.hpp"
 
 using namespace std::literals;
--- a/src/condition/cmdi_detector.cpp
+++ b/src/condition/cmdi_detector.cpp
@@ -357,7 +357,8 @@ std::string_view find_shell_command(std::string_view executable, object_view exe
 
 std::optional<shi_result> cmdi_impl(object_view exec_args,
     std::vector<shell_token> &resource_tokens, object_view params,
-    const object_set_ref &objects_excluded, ddwaf::timer &deadline)
+    std::span<const transformer_id> transformers, const object_set_ref &objects_excluded,
+    ddwaf::timer &deadline)
 {
     const std::string_view executable =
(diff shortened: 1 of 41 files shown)


Taiki-San (approved):
Fixing the TODO would be nice, since double decoding can result in broken outputs

src/condition/cmdi_detector.cpp
```diff
@@ -390,8 +393,15 @@ std::optional<shi_result> cmdi_impl(object_view exec_args,
         }
 
         if (eval_executable) {
-            // First check if the entire executable was injected
             auto value = param.as<std::string_view>();
+
+            // TODO: transforms are done twice if there is a shell command
```
cataphract, line 398: isn't avoiding this just a matter of passing the transformed value to find_shi_from_params, together with an empty list of transformers?

src/condition/lfi_detector.cpp
```diff
@@ -114,6 +118,17 @@ lfi_result lfi_impl(std::string_view path, object_view params,
             continue;
         }
 
+        auto transformed = transformer::manager::transform(param, transformers);
```
cataphract, line 121: i find it a bit awkward that every implementation needs to do this. what if instead passing `params` the _impl methods were given already lazy transforming iterators?

cataphract (approved):

src/condition/cmdi_detector.cpp
Anilm3: `find_shi_from_params` takes an object, on the other hand transformed is a `cow_string` which can't be converted into an `owned_object` as the allocated memory may not match it's length which is a restriction we have on objects. I'm looking to see if I can figure out a simple way to refactor `find_shi_from_params`.

src/condition/lfi_detector.cpp
Anilm3: Not a bad idea, I'll give it a go

PR #508, opened 2026-07-30
Improvements on RASP operator transformer support

Anilm3:
This PR makes the following improvements:

- Added `transforming_iterator`, an iterator adapter which yields transformed string values. This is used by lfi, cmdi and `match_iterator`.
- Removed the double transform in `cmdi_impl`, parameters are now transformed once instead of once per check.
- `match_iterator` refactored to move most logic to `transforming_iterator`.
- Added deadline to both iterators, given the newly introduced costs.
- Fixed a empty string transform resulting in a null buffer.
- Added regression tests for both fixes across all five detectors, plus unit tests for the new iterator.

### DIFF
--- a/src/condition/transforming_iterator.hpp
+++ b/src/condition/transforming_iterator.hpp
@@ -1,0 +1,123 @@
+// Unless explicitly stated otherwise all files in this repository are
+// dual-licensed under the Apache-2.0 License or BSD-3-Clause License.
+//
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2021 Datadog, Inc.
+
+#pragma once
+
+#include "clock.hpp"
+#include "cow_string.hpp"
+#include "exception.hpp"
+#include "exclusion/common.hpp"
+#include "iterator.hpp"
+#include "object.hpp"
+#include "transformer/base.hpp"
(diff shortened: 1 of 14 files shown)


Taiki-San (approved):

src/condition/transforming_iterator.hpp
```diff
@@ -0,0 +1,123 @@
...
+        while (++it_) {
+            if (deadline_.expired()) {
+                throw ddwaf::timeout_exception();
+            }
+
+            if (transform_current_object(*it_)) {
+                return true;
+            }
+        }
+        return false;
```
cataphract: similar to the constructor. perhaps extract

```cpp
bool advance_to_transformable_object()
{
    for (; it_; ++it_) {
        // Objects which can't be transformed are skipped within this
        // iterator, so the deadline must be evaluated here rather than
        // relying on the caller doing so once per yielded value
        if (deadline_.expired()) {
            throw ddwaf::timeout_exception();
        }

        if (transform_current_object(*it_)) {
            return true;
        }
    }
    return false;
}
```

the constructor would just call it; operator++ needs ++it_ before the call

cataphract (approved):

PR #509, opened 2026-07-31
Upload ELF debug symbols on release

florentinl:
## Summary

- add a tag-only job after the draft release is created
- download Linux musl packages and extract ELF `.debug` files, excluding unsupported i386
- mint short-lived Org 2 credentials with dd-sts and upload via `datadog-ci`

## Why

Publishing release debug symbols lets Datadog symbolize profiles that contain libddwaf frames without storing a long-lived Datadog API key in GitHub.

## Implementation

- Node 26, floating within the major version
- `@datadog/datadog-ci` 5.21.2, the current npm release
- latest stable action revisions pinned to full commit hashes: `download-artifact` v8.0.1, `setup-node` v7.0.0, and `dd-sts-action` v1.0.5
- upload x86_64, ARMv7, and AArch64 symbols; exclude i386 because `datadog-ci` does not support it

## Validation

- actionlint 1.7.12 schema and expression validation
- `yq` parsing and `git diff --check`
- Linux `datadog-ci --dry-run` against the three supported libddwaf 2.0.1 release archives: 3/3 symbols handled successfully

Companion dd-source policy PR: https://github.com/ddoghq/dd-source/pull/39414

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -401,3 +401,55 @@ jobs:
             ./artifacts/**/*.sha256
             ./artifacts/**/*.nupkg
             ./artifacts/**/*.json
+
+  upload-elf-debug-symbols:
+    needs: [release]
+    runs-on: ubuntu-24.04
+    if: startsWith(github.ref, 'refs/tags/')
+    permissions:
+      contents: read
+      id-token: write
+    steps:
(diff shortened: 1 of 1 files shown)


Anilm3 (approved):
lgtm

PR #510, opened 2026-07-31
Add option to convert strings on non-string matchers

Anilm3:


src/utils.cpp
```diff
@@ -100,6 +100,23 @@ template <typename T> std::pair<bool, T> from_string(std::string_view str)
...
+    }
+
+    if (str.size() == (sizeof("false") - 1) && (str[0] == 'F' || str[0] == 'f') &&
+        (str[1] == 'A' || str[1] == 'a') && (str[2] == 'L' || str[2] == 'l') &&
+        (str[3] == 'S' || str[3] == 's') && (str[4] == 'E' || str[4] == 'e')) {
+        return {true, false};
+    }
+
+    return {false, false};
+}
```
cataphract, line 118: I think you want to return something like std::optional<bool>

src/matcher/base.hpp
```diff
@@ -42,14 +51,15 @@ class base {
...
-    [[nodiscard]] virtual std::pair<bool, dynamic_string> match(object_view obj) const = 0;
+    [[nodiscard]] virtual std::pair<match_result, dynamic_string> match(
+        std::string_view str) const = 0;
+    [[nodiscard]] virtual std::pair<match_result, dynamic_string> match(object_view obj) const = 0;
 };
 
 template <typename T> class base_impl : public base {
 public:
     base_impl() = default;
+    explicit base_impl(bool convert_strings) : convert_strings_(convert_strings) {}
```
cataphract, line 62: whether to convert strings could be a non-type template parameter

cataphract (approved):

PR #511, opened 2026-08-24
Add large array and map representations

cataphract:
Store large-array and large-map metadata in the object itself while preserving
the 16-byte layout and the element-pointer offset shared with compact
containers.

## Representation

Large containers use the first 64 bits of the object for:

- an 8-bit type
- a 28-bit size
- a 28-bit capacity

The second 64 bits holds the direct element pointer. This removes the private
allocation headers and their flexible arrays; large-container allocations now
contain only elements and use the element type's alignment.

## Capacity limits

The maximum capacities are compile-time constants and are validated before
allocation:

- arrays: `min(268,435,455, SIZE_MAX / sizeof(ddwaf_object))`
- maps: `min(268,435,455, SIZE_MAX / sizeof(ddwaf_object_kv))`

On 32-bit platforms this allows 268,435,455 array elements and 134,217,727 map
entries. On wider platforms both limits are 268,435,455. Oversized creation and
exhausted growth fail without modifying the existing object.

### DIFF
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -173,37 +173,22 @@ struct _ddwaf_object_map {
 
 // Large-container metadata uses 28-bit fields for size and capacity, for a
 // maximum of 268,435,455 elements (potentially lower on 32-bit platforms due
-// to allocation-size limits). metadata._type overlays the direct uint8_t type
-// view, and ptr points directly to the contiguous elements. Any insertion that
-// grows the container invalidates ptr and pointers to its elements.
+// to allocation-size limits). _type overlays the direct uint8_t type view of
+// ddwaf_object, and ptr points directly to the contiguous elements. Any
+// insertion that grows the container invalidates ptr and pointers to its
+// elements.
 struct _ddwaf_object_large_array {
-    union {
-        uint8_t type;
-        struct {
-            uint64_t _type : 8;
-            uint64_t size : 28;
-            uint64_t capacity : 28;
-        } metadata;
-    };
-    union {
-        uint64_t _padding;
-        union _ddwaf_object *ptr;
-    };
+    uint64_t _type : 8;
+    uint64_t size : 28;
+    uint64_t capacity : 28;
+    union _ddwaf_object *ptr;
 };
 
 struct _ddwaf_object_large_map {
-    union {
-        uint8_t type;
-        struct {
-            uint64_t _type : 8;
-            uint64_t size : 28;
-            uint64_t capacity : 28;
-        } metadata;
-    };
-    union {
-        uint64_t _padding;
-        struct _ddwaf_object_kv *ptr;
-    };
+    uint64_t _type : 8;
+    uint64_t size : 28;
+    uint64_t capacity : 28;
+    struct _ddwaf_object_kv *ptr;
 };
 
 /**
--- a/include/ddwaf.h
+++ b/include/ddwaf.h
@@ -171,12 +171,6 @@ struct _ddwaf_object_map {
     struct _ddwaf_object_kv *ptr;
 };
 
-// Large-container metadata uses 28-bit fields for size and capacity, for a
-// maximum of 268,435,455 elements (potentially lower on 32-bit platforms due
-// to allocation-size limits). _type overlays the direct uint8_t type view of
-// ddwaf_object, and ptr points directly to the contiguous elements. Any
-// insertion that grows the container invalidates ptr and pointers to its
-// elements.
 struct _ddwaf_object_large_array {
     uint64_t _type : 8;
     uint64_t size : 28;
@@ -936,11 +930,7 @@ ddwaf_object* ddwaf_object_set_bool(ddwaf_object *object, bool value);
 ddwaf_object* ddwaf_object_set_float(ddwaf_object *object, double value);
 
 /**
- * Creates an array object for sequential storage. Capacities up to 65,535 use
- * DDWAF_OBJ_ARRAY; larger capacities use DDWAF_OBJ_LARGE_ARRAY. A compact array
- * is promoted automatically when insertion grows it beyond 65,535 elements.
- * Large arrays are limited to the smaller of 268,435,455 elements and
- * SIZE_MAX / sizeof(ddwaf_object).
+ * Creates an array object for sequential storage.
  *
  * @param object Object to perform the operation on. (nonnull)
  * @param capacity Initial capacity of the array.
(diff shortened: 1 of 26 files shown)


include/ddwaf.h
```diff
@@ -167,6 +171,26 @@ struct _ddwaf_object_map {
     struct _ddwaf_object_kv *ptr;
 };
 
+// Large-container metadata uses 28-bit fields for size and capacity, for a
+// maximum of 268,435,455 elements (potentially lower on 32-bit platforms due
+// to allocation-size limits). _type overlays the direct uint8_t type view of
+// ddwaf_object, and ptr points directly to the contiguous elements. Any
+// insertion that grows the container invalidates ptr and pointers to its
+// elements.
```
Anilm3:
```suggestion
// to allocation-size limits).
```

Looks like the typical prompt-leak within comments... This isn't particularly useful.

include/ddwaf.h
```diff
@@ -167,6 +171,26 @@ struct _ddwaf_object_map {
...
+struct _ddwaf_object_large_array {
+    uint64_t _type : 8;
```
Anilm3, line 175: `type` was fine, although since it's a `uint64_t` I guess it makes some sense to "hide" it

include/ddwaf.h
```diff
@@ -919,26 +936,34 @@ ddwaf_object* ddwaf_object_set_bool(ddwaf_object *object, bool value);
...
+ * SIZE_MAX / sizeof(ddwaf_object).
  *
  * @param object Object to perform the operation on. (nonnull)
  * @param capacity Initial capacity of the array.
  * @param alloc Allocator to use for memory allocation. (nonnull)
  *
  * @return A pointer to the passed object or NULL if the operation failed.
  **/
-ddwaf_object* ddwaf_object_set_array(ddwaf_object *object, uint16_t capacity, ddwaf_allocator alloc);
+ddwaf_object* ddwaf_object_set_array(ddwaf_object *object, size_t capacity, ddwaf_allocator alloc);
```
Anilm3: Unfortunately breaks the ABI and would require a major release.

src/interface.cpp
```diff
@@ -115,6 +117,27 @@ static_assert(offsetof(detail::object_map, size) == offsetof(_ddwaf_object_map,
...
 
+// detail::object_large_array == _ddwaf_object_large_array
+static_assert(sizeof(detail::object_large_array) == sizeof(_ddwaf_object_large_array));
+static_assert(
+    offsetof(detail::object_large_array, ptr) == offsetof(_ddwaf_object_large_array, ptr));
+static_assert(std::is_same_v<decltype(detail::object_large_array::ptr), detail::object *>);
+static_assert(std::is_same_v<decltype(_ddwaf_object_large_array::ptr), ddwaf_object *>);
+#if defined(__GNUC__) && !defined(__clang__)
+constexpr _ddwaf_object_large_array ddwaf_object_large_array_type_byte_probe{._type = 0xFF};
+static_assert(__builtin_memcmp(&ddwaf_object_large_array_type_byte_probe, "\xFF", 1) == 0);
```
Anilm3, line 131: What are you trying to prove with this?

src/object.hpp
```diff
@@ -122,13 +147,156 @@ static_assert(offsetof(object, type) == offsetof(object_small_string, type));
...
+    return size - static_cast<std::size_t>(distance);
+}
+
+constexpr std::size_t large_container_metadata_capacity_limit = (uint32_t{1} << 28) - 1;
+constexpr std::size_t max_large_array_capacity =
+    std::min(large_container_metadata_capacity_limit, maxof_v<std::size_t> / sizeof(object));
+constexpr std::size_t max_large_map_capacity =
+    std::min(large_container_metadata_capacity_limit, maxof_v<std::size_t> / sizeof(object_kv));
+
+static_assert(max_large_array_capacity == large_container_metadata_capacity_limit);
```
Anilm3:
```suggestion
constexpr std::size_t large_container_capacity_limit = (uint32_t{1} << 28) - 1;
constexpr std::size_t max_large_array_capacity =
    std::min(large_container_capacity_limit, maxof_v<std::size_t> / sizeof(object));
constexpr std::size_t max_large_map_capacity =
    std::min(large_container_capacity_limit, maxof_v<std::size_t> / sizeof(object_kv));

static_assert(max_large_array_capacity == large_container_capacity_limit);
```

src/object.hpp
```diff
@@ -122,13 +147,156 @@ static_assert(offsetof(object, type) == offsetof(object_small_string, type));
...
+}
+
+template <typename Element> inline const Element *container_data(const object &obj) noexcept
+{
+    Element *data;
+    // NOLINTNEXTLINE
+    memcpy(&data, reinterpret_cast<const std::byte *>(&obj) + offsetof(object_array, ptr),
+        sizeof(data)); // NOLINT(bugprone-sizeof-expression)
+    return data;
+}
```
Anilm3, line 270: I'm not sure I understand the point of these functions, when would the union member be inactive and what prompted their creation?

src/object.hpp
```diff
@@ -175,6 +343,111 @@ inline std::pair<T *, SizeType> realloc_helper(
...
+};
+
+template <> struct large_container_traits<object_kv> {
+    static constexpr std::size_t capacity_limit = max_large_map_capacity;
+};
+
+template <typename Element> inline std::size_t large_container_bytes(std::size_t capacity)
+{
+    if (capacity > large_container_traits<Element>::capacity_limit) [[unlikely]] {
+        throw std::length_error("container capacity exceeds allocation limit");
```
Anilm3: Should still be bad alloc if it's an allocation error.

src/object.hpp
```diff
@@ -175,6 +343,111 @@ inline std::pair<T *, SizeType> realloc_helper(
...
+
+template <typename Element>
+inline Element *alloc_large_container(std::size_t capacity, memory::memory_resource &alloc)
+{
+    if (capacity == 0) {
+        return nullptr;
+    }
+
+    const auto bytes = large_container_bytes<Element>(capacity);
+    auto *data = static_cast<Element *>(alloc.allocate(bytes, alignof(Element)));
```
Anilm3: `memory_resource::allocate` doesn't return `nullptr`, but rather directly throws `std::bad_alloc`.

src/interface.cpp
cataphract: that the first byte is used to store the type. This layout of bitfields is not guaranteed by the standard afaik

src/object.hpp
cataphract: I'm doing away with it. The compiler is smart enough to remove the test. On arm64 it becomes just `ldr x0, [x0, #8]`

include/ddwaf.h
cataphract: Yes, I'd prefer to avoid accidental uses of this member.

include/ddwaf.h
cataphract: As discussed, kept ddwaf_object_set_array symbol for ABI compatibility and redirected ddwaf_object_set_array to ddwaf_object_set_array_large. ddwaf_object_set_array is poisoned to avoid it being used directly

Anilm3 (approved):

PR #512, opened 2026-08-31
Add ddwaf_object_from_truncated_json (RFC-1089)

florentinl:
## Summary

Implements RFC-1089: `ddwaf_object_from_truncated_json()`, a parser that accepts JSON prefixes — inputs cut off partway through an otherwise valid document — by retaining completed values, finalizing open containers, and truncating open strings to their last complete escape sequence and UTF-8 code point.

Also hardens `ddwaf_object_from_json()` (embedded NUL, trailing content, null allocator).

> [!IMPORTANT]
> **Stacked on #511** (`glopes/large-repr`): the depth/container-saturation boundary tests rely on its large-container representations. The base branch is set accordingly; retarget to `master` once #511 lands.

### Rejection rules for truncated inputs

- **Syntax errors before the end of the input** — recovery only applies to truncation at the very end
- **Embedded NUL bytes** — the stream reports EOF as `'\0'`, so a mid-input NUL is indistinguishable from a short read
- **Containers nested more than 20 levels deep** (hard failure rather than silent drop, as `ddwaf_object_destroy` is still recursive)
- **Impossible escape prefixes** — e.g. `"\uDC`, which can only complete into a lone low surrogate; recoverable prefixes such as `"\uD83D\uDE0` are accepted
- **Terminal numbers** retained only if the digits present already form a valid JSON number (`[12` → `[12]`, `[12.` → `[]`)

### Validation gaps closed (found in review)

| # | Finding | Fix |
|---|---------|-----|
| 1 | Container overflow at 65,536 elements | Fixed by #511's promoted representations; regression-tested at 65,535/65,536/65,537 and verified under ASan |
| 2 | Embedded NUL / trailing content accepted | Consumed-length invariant after successful parse |
| 3 | Unsound surrogate recovery (`"\uDC` accepted) | Range-intersection semantics on the digits present |
| 4 | Silent depth drop in truncated mode | Parse termination → rejection; strict mode unchanged |
| 5 | Docs referenced `ddwaf_object_free`; restrictions unexplained | `ddwaf.h` notes rewritten; `docs/c-api/api.md` regenerated (idempotent) |
| 6 | Truncated tests matched unordered maps | New `EXPECT_JSON_EXACT` matcher (ordered arrays, exact integrality) |
| 7 | No fuzzing coverage | New `truncated_json` target, wired into CI (build failure fatal) |
| 8 | Null allocator dereference in `object_from_json` | Explicit rejection in both entry points |

## Testing

- **1679/1679 tests pass**, including under an ASan-instrumented build with zero reports
- New unit/integration tests: terminal numbers, embedded NUL, surrogate prefixes, depth-20 boundary, large containers, UTF-8 validation divergence (strict accepts invalid UTF-8 bytes, truncated validates encoding — an intentional divergence needed to block truncated `\xED` sequences from recovering into surrogates), null allocators
- Fuzzer asserts three invariants: (1) strict-valid ⇒ truncated-valid (ASCII inputs only, for the UTF-8 reason above), (2) NUL-containing inputs always rejected, (3) prefix-closure — every root-reaching prefix of a valid container/string input is valid
- The invariants were validated locally over 19 corpus seeds, every prefix of 3 valid documents, and 200,000 mutations (libFuzzer itself cannot link on macOS due to `-fsanitize=leak`; CI will exercise the real target)

## Notes for reviewers

- Strict-mode deep-nesting behavior (`NestedEmptyArraysHardcodedLimit/Maps`, silent drop) is intentionally unchanged
- `libddwaf.version` uses a `ddwaf_*` wildcard; only `libddwaf.def` needed the new export
- clang-tidy is unavailable locally; CI covers it

### Why the truncated mode hard-fails at depth 20

Depth in truncated input is attacker-controlled (a few hundred KB of `[[[[...` is ~200k nesting levels), and the object tree is consumed recursively (`object_destroy`, `object_clone` — the recursive calls are not in tail position since the container payload is freed after its children, so no TCO is possible). Unbounded depth would therefore turn a stack overflow into the failure mode instead of a clean rejection. Lifting or making the cap configurable is follow-up work: iterative destroy plus an audit of the other recursive traversals.

### Semantics pinned by implementation (worth confirming against the RFC text)

- Incomplete surrogate pairs are dropped as a whole: `["hello\uD83D\uDE0` recovers to `["hello"]` (the high surrogate is not retained on its own)
- Terminal numbers are retained (`["123` → `[123]`), matching `kParseStopWhenDoneFlag`-style prefix semantics
- `ddwaf_object_from_json` itself now rejects embedded NULs / trailing content / null allocators (bugfix-class behaviour change bundled here — can be split into a separate PR if preferred)


Anilm3:
@florentinl  I've not done a full review yet but why add a new function instead of just supporting truncated JSON on the existing one?

cataphract:
This adapts a function whose purpose is to parse trusted controlled input into one that's used to parse user input. In the latter case you would want to be more permissive. rapidjson has several flags to tolerate input that's sometimes deemed valid, like comments or single quotes (IIRC). There's also the recursion you mentioned. I'm also not a big fan of this method of truncation, I think what should be limited should be output not the input, o/wise it's trivial to bypass by just adding a bunch of whitespace, but that's another matter...

florentinl:
@cataphract the PR is far from ready for many reasons that you mentioned.
The PR is not yet open for review (hence the draft status). Still thanks for giving it a quick look.

> This adapts a function whose purpose is to parse trusted controlled input into one that's used to parse user input. In the latter case you would want to be more permissive. rapidjson has several flags to tolerate input that's sometimes deemed valid, like comments or single quotes (IIRC). 

I wasn't aware tracers are supposed to handle JSON with comments or other variants of JSON. We don't in python at least. I assume it must be an ecosystem thing.

> There's also the recursion you mentioned. 

Agreed, I think it would need to be fixed before this such a json helper lands.

>  I'm also not a big fan of this method of truncation, I think what should be limited should be output not the input, o/wise it's trivial to bypass by just adding a bunch of whitespace, but that's another matter...

It is still time to question the RFC. No tracer has implemented it as far as I am aware.

